PR #9, opened 2021-09-11
feat: Support starting p2p sessions after bevy has started

johanhelsing:
To allow things like lobbies, main menus etc.


johanhelsing:
This builds on #8. I can rebase it without it if you want

PR #14, opened 2021-10-03
Example: add checksum example for non reflect types

JonathanPicques:
This PR adds an example to compute a checksum for types that don't support reflect/hashing (floats, bevy::Transform, ...)

This checksum can be checked by synctest to detect mismatches


gschup (changes requested):
Love it! But I would prefer the usage of the "proper" rand crate instead of the extern C rand(), as example code is often used as a starting point for less experienced users.

gschup (approved):
great! thanks for the work!

PR #22, opened 2022-04-18
Bevy 0.7 + warning/clippy fixes

johanhelsing:
Haven't tested this yet, but think it should work.

I'm trying to port my achtung clone to 0.7. Hopefully I'll know soon enough whether it works.


johanhelsing:
Ok, I got around to fixing the examples. Synctest and p2p are working, haven't tried spectators

PR #25, opened 2022-07-30
updated for 0.8

paxsethorld:


Cargo.toml
```diff
@@ -16,7 +16,7 @@ categories = ["network-programming", "game-development"]
 wasm-bindgen = ["instant/wasm-bindgen", "ggrs/wasm-bindgen"]
 
 [dependencies]
-bevy = { version = "0.7", default-features = false }
+bevy = { version = "0.8.0", default-features = false, features = ["render", "bevy_asset","bevy_scene",]}
```
johanhelsing, line 19: Why do these need to be added to the dependencies? Can't they just be dev dependencies?

src/world_snapshot.rs
```diff
@@ -158,24 +158,24 @@ impl WorldSnapshot {
...
-                        None => reflect_component.remove_component(world, entity),
+                        None => reflect_component.remove(world, entity),
                     }
                 } else {
                     // the entity in the world has no such component
                     if let Some(component) = rollback_entity
                         .components
                         .iter()
-                        .find(|comp| comp.type_name() == registration.name())
+                        .find(|comp| comp.type_name() == registration.short_name())
```
johanhelsing, line 175: What's the difference between `short_name` and `type_name`? I tried finding the documentation for short_name for 0.8, but it appears to be missing (https://docs.rs/bevy/latest/bevy/reflect/struct.TypeRegistration.html#method.short_name), I just get a 404.

The documentation for `short_name` for 0.7, however, states that it's the type name with all paths removed.

What I'm worried is that if you have a type named `Translation`, `Visibility` etc. in your own crate, you might now get naming collisions? (Just want to make sure were not accidentally switching to the wrong name variant).

Cargo.toml
paxsethorld: Yes, should work fine as dev dependencies.

src/world_snapshot.rs
paxsethorld: type_name is the return value of std::any::type_name, while short_name is computed from bevy::utils::get_short_name

yes you're right, this may be a problem but other than naming collisions, updated version seems to working great with my game. Thanks so much for such a useful crate!

gschup:
Currently in vacation, but will take care of this later next week. Thanks for the PR :)

PR #26, opened 2022-08-10
update for 0.8 with some more changes

thwischm:
I tried to get #25 working and ran into some more issues. I don't know how to add more changes to someone else's PR, so here's a new one.

Apparently a bevy 0.8 TypeRegistry contains some primitive types by default, which caused a "Unregistered Type in GGRS Type Registry" panic. I couldn't find a convenient way to construct a TypeRegistryArc that's actually empty in the new docs, so we have to construct one directly?

I also had into short_name and type_name on a TypeRegistration actually returning different names for the same type. This made write_to_world overlook the components that were saved in the snapshot and delete them from the world instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ bytemuck = { version = "1.7", features=["derive"]}
 instant = "0.1"
 log = "0.4"
 ggrs = { git = "https://github.com/gschup/ggrs", features=["sync-send"]}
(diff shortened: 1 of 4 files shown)


johanhelsing:
Tested with my one of my games and seems to work great now :)

gschup:
Wonderful work :) thank you for the PR!

PR #28, opened 2022-11-06
Only Depend on Bevy `bevy_render` not `render`

zicklag:
This removes this crate's dependency on the Bevy renderers such as `bevy_text`/`ui`/`pbr`, etc.

It definitely was a little unintuitive that the `render` feature is different than the `bevy_render` feature.


gschup:
Nice catch! Thanks for the PR!

PR #30, opened 2022-11-08
Re-export `ggrs` Crate

zicklag:
This makes it easier to make sure your version of `ggrs` is the one compatible with `bevy_ggrs` and saves you from adding an extra dependency to your `Cargo.toml`.


gschup:
I learned something new today. Thanks for the PR!

PR #31, opened 2022-11-10
Map Entity IDs When Writing Snapshot to World

### LINKED ISSUE #29: [Discussion] Invalidated Entity's
To answer your side question:

GGRS rewinds by loading an old state and overwriting current values. When loading a snapshot, items that have been deleted will have to be recreated and thus get a new entity id. In the same time, your inventory will be overwritten with the entity id vector it had at that point in time. These IDs will then probably be invalid.

_Originally posted by @gschup in https://github.com/gschup/bevy_ggrs/issues/11#issuecomment-1306735337_
      
---

I started a new thread to avoid cluttering the other one.

So, if the entities in an inventory component could end up invalidated by the rollback, does that mean Bevy's hierarchy could get corrupted too?

Because the hierarchy works by having `Parent` and `Children` components where `Entity`s are used for references, similar to an inventory.

zicklag:
Resolves #29.

I haven't really tested this yet, but this should fix entity mapping and hierarchy problems.

If this works it'd be good to have some documentation on how this effects types that store entities.

For instance, if you create your own `Inventory` component that needs to be synced, you must implement and reflect `MapEntities` for it, so that it will get the new entity IDs when the snapshots are applied.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,9 +7,9 @@ use bevy::{
 };
 use ggrs::{Config, PlayerHandle};
 use ggrs_stage::GGRSStage;
+use parking_lot::RwLock;
 use reflect_resource::ReflectResource;
 use std::sync::Arc;
-use parking_lot::RwLock;
 
 pub use ggrs;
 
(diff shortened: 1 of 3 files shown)


gschup:
This looks lovely, but it would be good if you could test it, as well. We could simply add a smaller cube as a child on top of the player cubes in the examples. What do you think?

zicklag:
Yeah, we should definitely test it, but I think it requires a pretty specific test.

I think to run into the issue, we actually need to add a child to an entity, then delete the child in a later frame ( specifically we must use the `Children` component to decide which entity to delete ), then have it roll back, which will invalidate the entity as it gets re-created during the rollback. Then we just need to make sure that during a sync test the entity gets deleted without being re-created after rolling back and forward.

Actually I think that's isolated enough we can add it as a test case. I'll see if I can put that in a test and verify that it fails before this PR and works after this PR.

gschup:
Perfect!

zicklag:
There you go! I wrote a test case and discovered a couple bugs.

src/lib.rs
```diff
@@ -87,7 +87,20 @@ impl<T: Config + Send + Sync> Default for GGRSPlugin<T> {
...
+                    // data may be used.
+                    //
+                    // While this is a little bit of a weird spot to register these, are the only
+                    // Bevy core types implementing `MapEntities`, so for now it's probably fine to
+                    // just manually register these here.
+                    //
+                    // The user can still register any custom types with `register_rollback_type()`.
+                    r.register::<Parent>();
+                    r.register::<Children>();
+                    r
```
zicklag, line 102: You'll want to look at this section here. I'm not sure if there's a better place to do this. The comment explains.

gschup: I think this makes sense.

gschup:
Awesome work! Thank you so much!

PR #33, opened 2022-11-13
Add a RollbackEventHook Trait

zicklag:
This allows you to register functions that will be called in response to GGRS messages.

I think I'm going to need this or something like it for my game, but I haven't tested it yet, so it's pending evaluation. I just wanted to get it up so you could look at it and give your thoughts on the idea.

### DIFF
--- a/src/ggrs_stage.rs
+++ b/src/ggrs_stage.rs
@@ -1,5 +1,5 @@
-use crate::{world_snapshot::WorldSnapshot, SessionType};
-use bevy::{prelude::*, reflect::TypeRegistry};
+use crate::{world_snapshot::WorldSnapshot, RollbackEventHook, SessionType};
+use bevy::{ecs::entity::EntityMap, prelude::*, reflect::TypeRegistry};
 use ggrs::{
     Config, GGRSError, GGRSRequest, GameStateCell, InputStatus, P2PSession, PlayerHandle,
     SessionState, SpectatorSession, SyncTestSession,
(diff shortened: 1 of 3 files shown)


zicklag:
Just tested this out and it appears to be working! I did this really quick without a lot of thought on the names, so it could probably use some more careful thought, which I'll try to put in later.

Another thing that might need re-visiting is the reasoning behind passing in the `frame` and the `max_prediction` values. My use was to figure out which snapshot slot to put save/load the data from, just like `bevy_ggrs` does internally.

Maybe it should just return a `slot_id` instead, so that I'm not doing the modulo on the two values myself.

---

Anyway, if you're curious, I used this to hook the JS/TS scripting engine in with the rollback in Jumpy, so that now both the Rust, and the JS code can both rollback and restore seamlessly.

It turned out pretty neat. :D I can't wait to really get the game working with GGRS.

gschup:
I like the idea in general! Let's think about the API for such hooks.

Looking at the commits, it seems like some entity mapping commits snuck into this PR. This needs to be cleaned up.

zicklag:
Just did a on `main` now that entity mapping is merged.

> Let's think about the API for such hooks.

The specifics of what I ended up needing in my use-case was:

- The `pre_save` hook ( `post_save` would have worked too ), to give my scripts a chance to snapshot their state. The only input I needed here was the "slot" to store my snapshot in, but I calculated that from the `frame % max_snapshots`.
- The `post_load` hook, to give my scripts a chance to restore the state. In this case I needed the "slot" for the snapshot, same as before, and additionally the `EntityMap` so I could map any entities that I stored in the JS scope.

I figured I'd add the other hooks too, because they could be useful in different circumstances.

Later today, if I get the chance, I'll do a more careful pass over the code to see if there's anything I can improve.

PR #34, opened 2022-11-15
Fix Component Apply in Certain Cases

zicklag:
Removes and re-inserts components during snapshot restore, instead of using `apply()` which does an "update" not a "replace" of the current component data.

For example, an apply() will do an in-place update such that apply an array to an array will add items to the array instead of completely replacing the current array with the new one.

In Jumpy this caused a difficult to trace bug when updating the `Children` component for the Bevy hierarchy, where setting children to an empty array, actually left the component un-modified.


gschup:
Wow, that must have been a tough find. Thank you :)

PR #37, opened 2022-11-17
Add Tracing Spans

zicklag:
This can be really useful!

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ log = "0.4"
 #ggrs = { version= "0.9.2", features=["sync-send"]}
 ggrs = { git = "https://github.com/gschup/ggrs", features=["sync-send"]}
 parking_lot = "0.12.1"
+tracing = "0.1.37"
 
 [dev-dependencies]
 structopt = "0.3"
--- a/src/ggrs_stage.rs
+++ b/src/ggrs_stage.rs
@@ -198,6 +198,7 @@ impl<T: Config> GGRSStage<T> {
         }
     }
 
+    #[tracing::instrument(name = "GGRS Snapshot World", skip(self, cell, world))]
     pub(crate) fn save_world(
         &mut self,
         cell: GameStateCell<T::State>,
@@ -217,6 +218,7 @@ impl<T: Config> GGRSStage<T> {
         self.snapshots[pos] = snapshot;
     }
 
+    #[tracing::instrument(name = "GGRS Restore World", skip(self, world))]
     pub(crate) fn load_world(&mut self, frame: i32, world: &mut World) {
         self.frame = frame;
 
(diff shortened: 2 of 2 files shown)


gschup:
While I think this is a super cool debugging tool, putting this as a non-optional dependency seems suboptimal for users.

gschup:
I wonder what the "correct" way of using this plugin as a library is. Do you have more experience with this? How do other libraries/plugins use this? With a feature flag?

cscorley:
It should be possible to expose this under a feature flag using `cfg_attr`, e.g.: 

```
#[cfg_attr(feature = "trace", tracing::instrument(name = "GGRS Advance Frame", fields(frame = self.frame), skip(self, inputs, world)))]
```

In bevy_ggrs Cargo.toml, I believe something like this (taking hints from how the ["bevy/trace"](https://github.com/bevyengine/bevy/blob/main/Cargo.toml#L76) flags accomplishes this):

```
[features]
trace = [ "tracing" ]

[dependencies]
tracing = { version = "0.1.37", optional = true }
...
```

Then it should be `cargo run --features bevy_ggrs/trace` for users.

I have only done this for logging bevy systems before, but have not tested with something like Perfetto.  Seems neat :smile:


Users could also enable conditionally in their own Cargo.toml to group many together with `cargo run --features trace` 

```
[features]
default = [],
trace = [ "bevy_ggrs/trace", "bevy/trace", ... ]
```

zicklag:
The tracing dependency is enabled by default in Bevy, and it's what powers the `info!`, `error!`, etc. logging macros.

I think the `tracing` feature in Bevy just causes bevy to emit the JSON trace files, and I don't think it actually changes what get's logged.

That's not to say that adding the tracing instrumentation to the methods doesn't add overhead, though, similar to using `info!` for logging, and I'm not sure if Bevy uses the `tracing` crate's attribute macros, so that might cause a slight increase in compile times from building the macros.

I'll take a look at how bevy does the logging for it's stages some time, because it seems like it'd be reasonable to copy how they did that, since it's essentially the same use case.

If we add the tracing spans at a low enough log level such as `debug` or `trace`, users can actually compile out any runtime cost they might have by building with the `tracing/release_max_level_info` flag, and they could do that with no effort on our part, since it's built into tracing.

---

Anyway, next step: see how Bevy does the tracing spans for their stages, and then we can probably just follow their convention, since we aren't going to be adding much extra overhead with our spans I don't think, if Bevy is already doing it for their stages.

johanhelsing:
This is how bevy_render does it: https://github.com/bevyengine/bevy/blob/de98850a3e706a23280702e4836142917fe9bf68/crates/bevy_render/src/lib.rs#L263..L266

We could probably just do the same thing? Add a `trace` feature and create the spans inside the functions?

Not sure it the feature should be default or not, though? Might be kind of hard to discover?

zicklag:
> We could probably just do the same thing? Add a trace feature and create the spans inside the functions?

Yeah, seems right.

It probably shouldn't be default. I guess for discoverability we'll just have to leave a note in the README and hope people find it. 🤷‍♂️

gschup:
closed for inactivity. feel free to reopen when applicable.

Issue #35, opened 2022-11-16
Event Synchronization


zicklag:
**Is your feature request related to a problem? Please describe.**
The current design of Bevy GGRS doesn't snapshot and restore Bevy events, so events that are handled in a frame may not get handled again if the frame is rolled back to.

**Describe the solution you'd like**
We might need a custom event resource type that is compatible with the rollback and snapshot system.

**Describe alternatives you've considered**
A quick attempt to snapshot the `Events<T>` resource from Bevy didn't turn out so well. `Events<T>` isn't `Clone` even if `T: Clone`, which presents challenges, and the exposed public API doesn't allow us enough access to reasonably snapshot and restore it. Even if we could, we'd have to have a way to reset the read counters on the `Local<ManualEventReader<T>>` storage that powers the `EventReader<T>` system param.

Essentially bevy's built-in event system probably just isn't suitable for snapshot and restore, and we need some alternative, or we just have to make this as a caveat and say that you have to use components for sending events or something like that.

**Additional context**

gschup:
A lot of bevy's internal features are not very snapshot-friendly. As an example, bevy's current states uses a driver function that has a `Local<bool>` making it impossible to capture states without essentially rewriting the feature yourself (Which we did in bevy jam 1). Doing this for every feature that bevy might have cannot realistically be maintained.

zicklag:
That's absolutely reasonable, ad in that case I'm totally fine closing this with a note in the README about how it's an unsupported feature.

gschup:
There should probably be a section for well-known pitfalls / bevy features not working as expected.

johanhelsing:
I just had some trouble with `Added` and `Changed` queries triggering too often, causing bevy_ecs_ldtk to respawn the level on every frame during synctest sessions. 

Perhaps we can try to avoid unnecessary change detection? #39

zicklag:
@johanhelsing I replied on #39. 

---

To get started on a list of things to beware of for the README ( just real quick as I can think of them ):

- Writing/Reading a component/resource in a rollback system without registering it
    - Reading input directly from bevy
- Using `Local<T>` parameters in rollback systems
    - Using any other system parameters that depend on `Local` internally.
    - Using events
- Entity IDs will change over rollbacks if entities get deleted:
    - This means if you need to store an entity in a custom component you must implement [`MapEntities`](https://docs.rs/bevy/latest/bevy/ecs/entity/trait.MapEntities.html).
[...]

PR #38, opened 2022-12-02
Fix PR Bevy 0.9 Update

Vrixyz:
My attempt to help with #36.

I wanted to target your branch but I fumbled a bit, I could make a PR to your fork @johanhelsing if you're interested.

What I added with last commit:

cargo test passing with bevy 0.9.1

- renamed  and split `register_rollback_type` in `register_rollback_component` and `register_rollback_resource`
- all examples now compile
- up dependency to bevy 0.9.1 to use `resources.iter()`
- remove `refect_resource` mod

## What I tested

- cargo test.
- run the examples (2 players + 1 spectator) on mac.


examples/box_game/box_game_p2p.rs
```diff
@@ -81,48 +82,62 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            match sess.as_ref() {
+                Session::P2PSession(s) => {
+                    let num_players = s.num_players() as usize;
+                    for i in 0..num_players {
+                        if let Ok(stats) = s.network_stats(i) {
+                            println!("NetworkStats for player {}: {:?}", i, stats);
+                        }
+                    }
                 }
+                _ => panic!("This examples focus on p2p."),
```
johanhelsing, line 140: nit:
```suggestion
                _ => panic!("This example focuses on p2p."),
```

examples/box_game/box_game_spectator.rs
```diff
@@ -60,45 +61,51 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    for event in session.events() {
-        println!("GGRS Event: {:?}", event);
+fn print_events_system(mut session: ResMut<Session<GGRSConfig>>) {
+    match session.as_mut() {
+        Session::P2PSession(s) => {
+            for event in s.events() {
+                println!("GGRS Event: {:?}", event);
+            }
+        }
+        _ => panic!("This examples focus on p2p."),
```
johanhelsing, line 90: ditto

src/lib.rs
```diff
@@ -5,34 +5,33 @@ use bevy::{
...
 use parking_lot::RwLock;
-use reflect_resource::ReflectResource;
+// use reflect_resource::ReflectResource;
 use std::sync::Arc;
 
 pub use ggrs;
 
 pub(crate) mod ggrs_stage;
-pub(crate) mod reflect_resource;
+// pub(crate) mod reflect_resource;
```
johanhelsing, line 17:
```suggestion
```

src/lib.rs
```diff
@@ -5,34 +5,33 @@ use bevy::{
     prelude::*,
     reflect::{FromType, GetTypeRegistration, TypeRegistry, TypeRegistryInternal},
 };
-use ggrs::{Config, PlayerHandle};
+use ggrs::{Config, InputStatus, P2PSession, PlayerHandle, SpectatorSession, SyncTestSession};
 use ggrs_stage::GGRSStage;
 use parking_lot::RwLock;
-use reflect_resource::ReflectResource;
+// use reflect_resource::ReflectResource;
```
johanhelsing, line 11:
```suggestion
```

src/lib.rs
```diff
@@ -5,34 +5,33 @@ use bevy::{
...
+    P2PSession(P2PSession<T>),
+    SpectatorSession(SpectatorSession<T>),
 }
 
-impl Default for SessionType {
-    fn default() -> Self {
-        SessionType::SyncTestSession
-    }
-}
+// TODO: more specific name to avoid conflicts?
```
johanhelsing, line 32: I think we should probably think about what to name this... It's public API that will be used, and `Inputs` quite easily conflicts with something else... and is very similar to Bevy's `Input`.

gschup:
thank you for working on this! The two-player example seems to work, but the spectator example crashes immediately. 

```
cargo run --example box_game_spectator -- --local-port 7002 --num-players 2 --host 127.0.0.1:7000
```
leads to:
```
thread 'main' panicked at 'This examples focus on p2p.', examples/box_game/box_game_spectator.rs:90:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\Georg Schuppe\.cargo\registry\src\github.com-1ecc6299db9ec823\bevy_tasks-0.9.1\src\task_pool.rs:273:45
```

src/lib.rs
```diff
@@ -55,7 +54,7 @@ impl Rollback {
 
 /// Provides unique ids for your Rollback components.
 /// When you add the GGRS Plugin, this should be available as a resource.
-#[derive(Default)]
+#[derive(Resource, Default)]
```
johanhelsing, line 57:
```suggestion
#[derive(Resource, Default)]
#[reflect(Resource)]
```

src/lib.rs
```diff
@@ -5,34 +5,33 @@ use bevy::{
...
+    SpectatorSession(SpectatorSession<T>),
 }
 
-impl Default for SessionType {
-    fn default() -> Self {
-        SessionType::SyncTestSession
-    }
-}
+// TODO: more specific name to avoid conflicts?
+#[derive(Resource, Deref, DerefMut)]
```
johanhelsing, line 33:
```suggestion
#[derive(Resource, Deref, DerefMut)]
#[reflect(Resource)]
```

src/world_snapshot.rs
```diff
@@ -1,12 +1,12 @@
...
     ecs::{entity::EntityMap, reflect::ReflectMapEntities},
     prelude::*,
-    reflect::{Reflect, TypeRegistry},
+    reflect::{Reflect, TypeData, TypeRegistry},
     utils::HashMap,
 };
 use std::{fmt::Debug, num::Wrapping};
 
-use crate::reflect_resource::ReflectResource;
+// use crate::reflect_resource::ReflectResource;
```
johanhelsing, line 9:
```suggestion
```

johanhelsing (changes requested):
Got to admit I don't understand all of the reflect stuff. But most of this looks sensible to me (though I wrote parts of it myself :P)

Added mostly nitpick comments.

As I understand it what were previously part of the GGRS `reflect_resource` module is now implemented within Bevy itself. I guess that means we should also delete the reflect_resource module?

You mentioned you didn't do much checking... One thing you could do, is check if the frame count resource in the p2p example works as expected?

src/lib.rs
Vrixyz: What does this accomplish ? wouldn't Resource already have some reflect capabilities ? or is it to enable bevy_egui reflection ?

johanhelsing: Hmm.. maybe it's not needed? I think I maybe just made a brainfart and thought you added `derive(Reflect)`.

I guess the real answer is to just check what works :P

src/lib.rs
Vrixyz: I think I agree, but at t he same time it's namespaced so it's not a very big issue. What do you think of `SessionInputs` OR `SessionFrameInput` ? Also this could use some type documentation.

examples/box_game/box_game_spectator.rs
Vrixyz: There's also a indication of the crash encountered by @gschup : this example is not focusing on "p2p" (well it is, but it's a spectator session...) so this sentence and the match is not correct.

examples/box_game/box_game.rs
```diff
@@ -32,7 +32,7 @@ impl Config for GGRSConfig {
 }
 
 #[repr(C)]
-#[derive(Copy, Clone, PartialEq, Pod, Zeroable)]
+#[derive(Copy, Clone, PartialEq, Eq, Pod, Zeroable)]
```
Vrixyz, line 35: `Eq` was not really necessary but it pleases clippy :)

examples/box_game/box_game.rs
```diff
@@ -50,9 +50,8 @@ pub struct Velocity {
     pub z: f32,
 }
 
-// You can also register resources. If your Component / Resource implements Hash, you can make use of `#[reflect(Hash)]`
-// in order to allow a GGRS `SyncTestSession` to construct a checksum for a world snapshot
-#[derive(Default, Reflect, Hash, Component)]
+// You can also register resources.
+#[derive(Resource, Default, Reflect, Hash)]
```
Vrixyz, line 54: I'm not sure if reflect/hash is still needed, I didn't check

johanhelsing: My understanding of this is still kind of shallow, but `register_rollback_resource` seems to need Reflect because it's needed for `TypeRegistration::insert`...

Or more specifically `TypeRegistration::insert` needs the `TypeData` trait to be implemented, which is automatically implemented by deriving `Reflect`.

I think `#[derive(Hash)]` and `#[reflect(Hash)` is not needed. At least the examples, including synctest, run fine without them. Perhaps @gschup would like to confirm?

src/lib.rs
johanhelsing: Yep, sorry about the noise. We could perhaps add reflect in another PR. It might be useful for debugging.

src/lib.rs
johanhelsing: yeah, it's easy enough to fix... More thinking about it being easy to read.

We have prefixed `GGRSError`, `GGRSConfig`. And some have `P2P` as prefix (`P2PSession`).

1. `GGRSInputs`
2. `P2PInputs`
3. `PlayerInputs` (matches `PlayerHandle` which is used to index this resource)
4. `SessionInputs`

Just spitballing. Don't really have a strong opinion. I'm also ok with `Inputs`. What do you think @gschup ?

definitely agree it could use some documentation!

examples/box_game/box_game_spectator.rs
```diff
@@ -60,45 +61,51 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    for event in session.events() {
-        println!("GGRS Event: {:?}", event);
+fn print_events_system(mut session: ResMut<Session<GGRSConfig>>) {
+    match session.as_mut() {
+        Session::SpectatorSession(s) => {
+            for event in s.events() {
+                println!("GGRS Event: {:?}", event);
+            }
+        }
+        _ => panic!("This examples focuses on spectator."),
```
johanhelsing, line 90:
```suggestion
        _ => panic!("This example focuses on spectators."),
```

examples/box_game/box_game_spectator.rs
```diff
@@ -60,45 +61,51 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         if let Some(sess) = p2p_session {
-            if let Ok(stats) = sess.network_stats() {
-                println!("NetworkStats : {:?}", stats);
+            match sess.as_ref() {
+                Session::SpectatorSession(s) => {
+                    if let Ok(stats) = s.network_stats() {
+                        println!("NetworkStats : {:?}", stats);
+                    }
+                }
+                _ => panic!("This examples focuses on p2p."),
```
johanhelsing, line 108:
```suggestion
                _ => panic!("This example focuses on spectators."),
```

examples/box_game/box_game_p2p.rs
```diff
@@ -81,48 +82,62 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    for event in session.events() {
-        println!("GGRS Event: {:?}", event);
+fn print_events_system(mut session: ResMut<Session<GGRSConfig>>) {
+    match session.as_mut() {
+        Session::P2PSession(s) => {
+            for event in s.events() {
+                println!("GGRS Event: {:?}", event);
+            }
+        }
+        _ => panic!("This examples focuses on p2p."),
```
johanhelsing, line 119:
```suggestion
        _ => panic!("This example focuses on p2p."),
```

johanhelsing (commented):
I tested resource rollback by adding some logging to the synctest session example's `FrameCount`. Everything works as expected :D

Also ported my tutorial to this branch, and it works fine (though that doesn't really need resource rollback): https://github.com/johanhelsing/extreme_bevy/tree/bevy-0.9-wip

Feels like this is really close now, just some minor nits and some docs for the new resources would be nice. Great work @Vrixyz !

src/lib.rs
gschup: properly naming the Inputs is tough, but I think we should definitely not name it `Inputs`. I like `PlayerInputs` quite a bit. I would shy away from naming it `GGRSInputs` as this struct does not come from the GGRS crate itself.

examples/box_game/box_game.rs
gschup: Deriving and reflecting hash is useful [here](https://github.com/Vrixyz/bevy_ggrs/blob/1cc28cce4d3b5e3216b893dc917fdacf7f37048b/src/world_snapshot.rs#L120). It is definitely optional.

johanhelsing: I guess we could be more specific about it in the comment... Something like "By deriving `Hash` and adding `reflect(Hash)` the resource can automatically be checked for desyncs during a `SyncTestSession`"?

gschup: There is definitely a lot to be improved regarding documentation (see e.g. #35). It might probably even be better to that outside of this PR, in a big documentation commit.

gschup:
I don't think we are missing much here. I think it would be nice to rename `Inputs` to `PlayerInputs` and then, we can merge :)

PR #42, opened 2023-01-10
Add rollback command extensions for reduced boilerplate

johanhelsing:
I created this mainly because I was annoyed by having to pass the `RollbackIdProvider` around everywhere, but another interesting side-effect is that if you use this API, you don't actually have to know about the `RollbackIdProvider` resource or `Rollback` component in order to use bevy_ggrs.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1,0 +1,47 @@
+use bevy::{
+    ecs::system::{Command, EntityCommands},
+    prelude::*,
+};
+
+use crate::{Rollback, RollbackIdProvider};
+
+pub trait RollbackCommandsExt<'w, 's> {
+    fn spawn_rollback<'a, T: Bundle>(&'a mut self, bundle: T) -> EntityCommands<'w, 's, 'a>;
+}
+
+impl<'w, 's> RollbackCommandsExt<'w, 's> for Commands<'w, 's> {
+    fn spawn_rollback<'a, T: Bundle>(&'a mut self, bundle: T) -> EntityCommands<'w, 's, 'a> {
+        let mut entity_commands = self.spawn(bundle);
+        entity_commands.insert_rollback();
+        entity_commands
+    }
+}
+
+pub trait RollbackEntityCommandsExt {
+    fn insert_rollback(&mut self) -> &mut Self;
+}
+
+impl<'w, 's, 'a> RollbackEntityCommandsExt for EntityCommands<'w, 's, 'a> {
+    fn insert_rollback(&mut self) -> &mut Self {
+        let entity = self.id();
+        self.commands().add(InsertRollback { entity });
+        self
+    }
+}
+
+struct InsertRollback {
+    entity: Entity,
+}
+
+impl Command for InsertRollback {
+    fn write(self, world: &mut World) {
+        let mut rip = world.resource_mut::<RollbackIdProvider>();
+
+        let insert = bevy::ecs::system::Insert {
+            entity: self.entity,
+            bundle: Rollback::new(rip.next_id()),
+        };
+
+        insert.write(world);
+    }
+}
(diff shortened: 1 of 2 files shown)


johanhelsing:
@zicklag maybe this is of interest to you as well?

johanhelsing:
Hmm... After all, I think maybe this is a footgun. I think perhaps we will lose system order ambiguity warnings?

zicklag:
That's an interesting idea. 🤔

I don't know that it actually exacerbates the ordering issue, because you still have the issue, even if you include the rollback ID provider resource in your system.

It does delay the generation of the rollback IDs until the command buffers are flushed, but I would hope that if you have a specified order for your systems, that their command buffers would be flushed in the specified order.

And if you haven't specified a system order, then you aren't really any worse off, because any system spawning rollback components could borrow the rollback ID provider in any order.

---

In other words, if explicitly ordered systems ( i.e. ones with `.before()` or .`after()` constraints ), have their command buffers flushed in the same order they are run, then I think this is totally fine, and nicer to use.

Maybe we can ask the bevy folks about the command flushing order.

johanhelsing:
I would *really* miss those warnings...

An alternate approach would be to encapsulate both Commands and RollbackIdProvider in a SystemParam. Would be a bit awkward if you need to access the inner Commands, though

zicklag:
> I would really miss those warnings...

Oh... I just realized what you're saying about the ambiguity warnings. I hadn't realized that Bevy can only give you those because of the conflicting mutable access to the resource. 

Yeah, that would be a big thing to use when determinism is critical to rollback even working right.

---

The system param is a good idea. You might be able to just deref to the inner commands, but at the same time it almost makes things more confusing. At least with the way it is now, it's really easy to tell what's going on.

Maybe we just need to make the rollback ID provider easier to use by having a method that actually returns `Rollback` components so you at least don't have to do the `Rollback::new(rollback_id_provider.next())` ( or whatever it is, I can't remember exactly ).

johanhelsing:
Okay, so one of the reason I started looking into this in the first place:

When making bundles, I was wondering whether to include the `Rollback` component or not. On one hand, it really seems like it makes sense to include when the bundle should be rolled back, on the other, it means the bundle can't implement `Default`, and it's really awkward for bundles with a lot of components, so I usually tend to spawn a bundle + a `Rollback` component, this means I have a lot of slightly awkward code like this:

```rust
fn spawn_player(mut commands: Commands, mut rip: ResMut<RollbackIdProvider>) {
    commands.spawn((
        CharacterBundle {
            name: Name::new("P1"),
            ..default()
        },
        Rollback::new(rip.next_id())
    ));
}
```

With the `SystemParam` approach, this would be:

```rust
fn spawn_player(mut commands: RollbackCommands) {
    commands.spawn_rollback(CharacterBundle {
        name: Name::new("P1"),
        ..default()
    });
}
```

Alternatively, it's probably possible to do something like this:

```rust
fn spawn_player(mut commands: Commands, mut rip: ResMut<RollbackIdProvider>) {
    commands.spawn(CharacterBundle {
        name: Name::new("P1"),
        ..CharacterBundle::with_rollback(Rollback::new(rip.next_id()))
    });
}
```

Or perhaps make a `RollbackBundle` trait and do something like this:

```rust
fn default_with_rollback<T: RollbackBundle>(rip: &mut RollbackIdProvider) -> T {
    T::default_with_rollback(&mut rip);
}

// instead of implementing default
impl RollbackBundle for CharacterBundle {
    fn default_with_rollback(mut rip: &mut RollbackIdProvider) {
        name: Name::new("Character"),
        // ... other "default" fields"
        rollback: rip.next()
    }
}

fn spawn_player(mut commands: Commands, mut rip: ResMut<RollbackIdProvider>) {
    commands.spawn(CharacterBundle {
        name: Name::new("P1"),
        ..default_with_rollback(&mut rip)
    });
}
```

Idk, maybe none of this is worth the extra complexity :/

In any case, I agree it would make things at least somewhat more convenient if we add a new method to `RollbackIdProvider` that returns a `Rollback`

PR #43, opened 2023-01-10
feat: Add `RollbackIdProvider::next`

johanhelsing:
As discussed in #42 .

While this slightly cuts down on the boilerplate, it also slightly obfuscates what type of component (`Rollback`) were adding...

What do you think?


gschup:
No matter on how we decide to continue with #42, this is a good little function. Before we merge this PR, could you please make the examples use this new change?

PR #44, opened 2023-01-10
feat: Add RollbackCommands system param

johanhelsing:
Alternative to #42


gschup:
is this still necessary?

johanhelsing:
Nah, probably too much magic

PR #48, opened 2023-02-09
Enable desync detection in p2p example

johanhelsing:
This is a very useful feature, we should either enable it by default, or at least inform about it in example code.

Unfortunately, I think it's currently broken? I get desyncs as soon as the game starts without giving any input.

@HeatXD


HeatXD:
Mm just tested it in the normal examples and it worked fine. i will look at this.

HeatXD:
> 

Does bevy's worlds have some non gameplay relevant state which is included in the snapshot?

johanhelsing:
seems they're using `AHasher::default` in bevy

johanhelsing:
Was able to make it work with this branch: https://github.com/johanhelsing/bevy/tree/fixed-reflect-hash-0.9.1

gschup:
very interesting! let's see how this goes :)

johanhelsing:
Tested it in Cargo Space as well, works great!

And I had no desyncs :tada: (or at least not any for my components that implement reflect_hash)

gschup:
is this PR still a draft?

johanhelsing:
Did a Ggrsevent use get lost in the merge?

johanhelsing:
Yeah, no problem! Just trying to help. Thanks for the super quick release!

PR #50, opened 2023-03-04
Bevy 0.10, simple port

### LINKED ISSUE #49: Sessions `.with_desync_detection(DesyncDetection::On)` always reports false positives
**Describe the bug**

Check out the branch in #48 , which enables desync detection in the p2p example. As soon as it starts, it reports a desync.

**Expected behavior**
It should not report a desync (since it doesn't desync. only the hash is different)

**Desktop (please complete the following information):**
 - Linux

**Additional context**
This happens because bevy_reflect uses `ahash::AHasher::default`. Since bevy uses ahasher with the default features, this means it's seeded at runtime, which means hashes will always be different across peers, see: https://docs.rs/ahash/latest/ahash/struct.AHasher.html#method.default

johanhelsing:
I know you plan a rewrite, I just really wanted to start playing around with bevy 0.10 and stageless.

This port is rather minimal; I changed as few lines as I could to get it working. There are probably a lot of things that could be simplified now, but wanted to check in first case you're already working on something :)

- [x] switch back to crates.io when 0.10 is actually released

Fixes #49

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ categories = ["network-programming", "game-development"]
 wasm-bindgen = ["instant/wasm-bindgen", "ggrs/wasm-bindgen"]
 
 [dependencies]
-bevy = { version = "0.9.1", default-features = false, features = ["bevy_render", "bevy_asset","bevy_scene",]}
+bevy = { git = "htps://github.com/bevyengine/bevy", default-features = false, features = ["bevy_render", "bevy_asset","bevy_scene",]}
 bytemuck = { version = "1.7", features=["derive"]}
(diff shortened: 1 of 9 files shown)


johanhelsing:
There's *a lot* more potential in Bevy 0.10, but I think this serves as an ok stepping stone for other prs :)

johanhelsing:
This is a very simple stupid port. I'd very much prefer if we merged #51 instead (or as well, iirc #51 builds on this one), it makes adding rollback systems a lot more ergonomic to use, and makes ambiguous system ordering in the rollback schedule an error (which I think would be a lifesaver for a lot of new users)

PR #59, opened 2023-06-15
Extension trait to add the GGRS plugin more idiomatically

NiklasEi:
The second commit contains some renaming based on [Rust's API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case). If you are set on the current naming I can revert it :slightly_smiling_face:


johanhelsing (commented):
I like this change, both the following naming conventions and the extension method.

Maybe also update the examples to use the extension?

NiklasEi:
Thanks, I had missed the `spectator` and `synctest` examples.

johanhelsing (approved):

gschup:
Thanks for the PR! This seems like a good idea!

PR #60, opened 2023-06-22
Fix add_rollback command extension

johanhelsing:
I think I just made a mistake in my comment here: https://github.com/gschup/bevy_ggrs/pull/58#discussion_r1173584216

We should just do the same as what `.insert` does.

This makes the following work:

```rust
        commands
            .entity(tilemap_entity)
            .insert(
                TilemapBundle {
                    grid_size,
                    map_type,
                    size: tilemap_size,
                    storage: tile_storage,
                    texture: TilemapTexture::Single(sprites.white_pixel.clone()),
                    tile_size,
                    transform: tilemap_transform.with_z(SHIP_BACKGROUND_LAYER_DEPTH),
                    ..default()
                }
            )
            .add_rollback();
```

### DIFF
--- a/src/rollback.rs
+++ b/src/rollback.rs
@@ -1,7 +1,10 @@
-use bevy::{prelude::{Component, Entity, World}, ecs::system::{EntityCommand, EntityCommands}};
+use bevy::{
+    ecs::system::{EntityCommand, EntityCommands},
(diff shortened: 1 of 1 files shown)


johanhelsing:
@bushrat011899 fyi

PR #62, opened 2023-09-11
Remove unneeded bevy features

johanhelsing:
Unless I'm doing something wrong, I don't think these are actually needed.


gschup:
I also don't think so. I forgot why they are there in the first place, to be honest. In any case, good that they are gone now :)

PR #64, opened 2023-09-28
Make `GgrsPlugin` a regular Bevy plugin

johanhelsing:
## Api changes

- `GgrsPlugin` is now just a regular bevy plugin, and the builder is gone.
- Adds a `ReadInputs` schedule, where inputs should be inserted into a `LocalInputs` resource.
- Registering rollback types are now done on the app instead.
- Adds `LocalPlayers` resource which contains the ggrs handles for the local players

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,11 +5,13 @@ use bevy::{
     ecs::schedule::{LogLevel, ScheduleBuildSettings, ScheduleLabel},
     prelude::*,
     reflect::{FromType, GetTypeRegistration, TypeRegistry, TypeRegistryInternal},
+    utils::HashMap,
 };
 use ggrs::{Config, InputStatus, P2PSession, PlayerHandle, SpectatorSession, SyncTestSession};
-use ggrs_stage::GgrsStage;
+use instant::{Duration, Instant};
 use parking_lot::RwLock;
-use std::sync::Arc;
+use std::{marker::PhantomData, sync::Arc};
+use world_snapshot::RollbackSnapshots;
 
 pub use ggrs;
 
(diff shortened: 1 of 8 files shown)


johanhelsing:
This is a more conservative alternative to #52 with the goal of just flattening the plugin (no Save/Load schedule).

The intention is that it should be easy to review (small diff) and focus on just one area, then we can continue with Save/Load schedules and non-reflect rollback in separate PRs.

gschup:
Is this a draft or is this ready for review?

johanhelsing:
> Is this a draft or is this ready for review?

Ready for review :)

src/lib.rs
```diff
@@ -110,56 +119,98 @@ impl<T: Config + Send + Sync> GgrsPlugin<T> {
...
-        let mut stage = GgrsStage::<T>::new(input_system);
-        stage.set_update_frequency(self.fps);
+// Label for the schedule which reads the inputs for the current frame
+#[derive(ScheduleLabel, Debug, Hash, PartialEq, Eq, Clone)]
+pub struct ReadInputs;
+
+/// GGRS plugin for bevy.
+pub struct GgrsPlugin<C: Config> {
+    /// phantom marker for ggrs config
+    _marker: PhantomData<C>,
```
gschup, line 143: Is this still the correct way to do this? In the past, I have seen this solution often, but it seems so clunky.

gschup (approved):

src/lib.rs
johanhelsing: Yeah, I don't think there's any way around it, unfortunately.

PR #65, opened 2023-10-13
Add more types to prelude and use it in examples

johanhelsing:
Adds app extensions to prelude, and a few `ggrs` re-exports.


gschup (approved):
Looks great!

PR #66, opened 2023-10-15
Add `SaveWorld`/`LoadWorld` schedules

johanhelsing:
Saving loading, same approach as in #52, but with checksums fixed.

Paves the way for more modular (e.g. `Clone`-based snapshotting).


johanhelsing:
I want to get something similar to [@bushrat011899's plugin-based snapshotting](https://github.com/gschup/bevy_ggrs/pull/52#issuecomment-1517198025). Just thought it's good to get this more uncontroversial prep step in first.

gschup (changes requested):
after merging #67 and #67, there now appears to be a merge conflict. I am sorry to inflict this on you, but it should be a simple fix 😅

gschup:
This is the part where I got a weird initialization bug in the rewrite branch. I don't remember a 100%, but the p2p example would sometimes crash due to some resources not being available (a schedule running before it should). For some reason, it happened only sometimes. We should check if the examples are working as intended.

johanhelsing:
I only ran the box_game p2p test once and synctest once, that worked. I can also give it a go in my own game for a more complex test.

johanhelsing:
> after merging #67 and #68, there now appears to be a merge conflict. I am sorry to inflict this on you, but it should be a simple fix 😅

No problem :)

gschup (approved):
All of this is in line with the ideas from #52, so I really cannot say anything. It also works on my machine. Thanks for doing this in a rigorous, orderly manner! Feel free to merge once you tried it with your game and are confident it works!

johanhelsing:
One difference between this and #52, is that in #52 run_ggrs_schedules is added to the the `Update` schedule instead of `PreUpdate`... I'm not sure what is more correct, or whether it's the source of any of any issues in #52, but just thought I'd mention.

johanhelsing:
> @johanhelsing do you have any remaining concerns with this PR? I think this is ready to be merged after fixing the merge conflicts.

Yeah, I was holding off because I was uncertain about the desync detection issues in #71 (but that turned out to just be the detection that was broken an irrelevant to this PR). I've tested it 
with extreme bevy (and that works great), but not in cargo space.

Will rebase and merge

PR #67, opened 2023-10-15
Use Bevy abstractions instead of using `instant` directly

johanhelsing:
This should make writing integration tests easier, since we can then use `TimeUpdateStrategy::ManualDuration`.


gschup (approved):
this makes a lot of sense!

gschup:
Learning moment for me: are there other bevy utils/abstractions we should use?

johanhelsing:
Not that I've seen so far. But yeah, I think it makes sense to use the bevy re-exports import paths where possible to make sure we don't use incompatible versions (or cause duplicates in the dependency tree).

PR #68, opened 2023-10-15
Fix panic when `set_rollback_schedule_fps` is not called

johanhelsing:
The resource was only added when modifying the FPS, so default FPS didn't work.

This was a regression in 57ceab38


gschup (approved):
good that you caught it!

Issue #54, opened 2023-03-30
Integration Tests incorrect


gschup:
I deleted [the integration tests](https://github.com/gschup/bevy_ggrs/blob/de326196cd484bbe74a0c0b7d25219fe304d7d19/tests/example_integration.rs) I previously merged. While I like the idea of having more testing not only on ggrs, but also the plugin, these tests did not correctly test what should happen. This should be fixed at some point.

donedgardo:
I'd love to contribute for this.
In general can we get a conversation going on what should be tested instead?

As a person integrating this library into their bevy game I'd like to see that:
* Systems in the Rollback stage should run.
* Rollback Components get synced between two applications with active session.
* Test simple integration (great for documentation) between bevy and the plugin.

The test in the link provided would be a good start, I would remove line https://github.com/gschup/bevy_ggrs/blob/de326196cd484bbe74a0c0b7d25219fe304d7d19/tests/example_integration.rs#L29
It is not expected that the two systems are completely in sync only that both system are being scheduled and executed.
Additionally the second test would be a good candidate for the integration test of rollback components: 
https://github.com/gschup/bevy_ggrs/blob/de326196cd484bbe74a0c0b7d25219fe304d7d19/tests/example_integration.rs#L35

gschup:
solved in #69 . we now have a nice basis to extend integration tests from if we wish to do so!

PR #69, opened 2023-10-15
Add integration tests

johanhelsing:
Depends on #66 I'll just rebase this after that one's in.


gschup:
#66 is in, so you are free to go :)

gschup (approved):
Very nice! This alleviates the issues I failed to properly explain in #54.

PR #73, opened 2023-10-19
Added Sensible Default `ggrs::Config` type `GgrsConfig`

bushrat011899:
# Objective

In the `box_game` example, there's a to-do related to `ggrs::Config`:

> `/// TODO: Find a way to hide the state type.`

## Solution

I've solved this to-do by creating a type `GgrsConfig`:

```rust
#[derive(Debug)]
pub struct GgrsConfig<Input, Address = SocketAddr, State = u8> {
    _phantom: PhantomData<(Input, Address, State)>,
}

impl<Input, Address, State> Config for GgrsConfig<Input, Address, State>
where
    Self: 'static,
    Input: Send + Sync + PartialEq + bytemuck::Pod,
    Address: Send + Sync + Debug + Hash + Eq + Clone,
    State: Send + Sync + Clone,
{
    type Input = Input;
    type State = State;
    type Address = Address;
}
```

Because `GgrsConfig` is a concrete type, it is allowed to have defaults for the type parameters `Input`, `State`, and `Address` required by `ggrs::Config`. As mentioned in the `box_game` example, `u8` for `State` is already taken as a default (since `State` is unused in Bevy). I also took the liberty of setting `Address` to `SocketAddr` as I believe that is the most common type that will be used by end users. These defaults allow the example game to use:

```rust
pub type BoxConfig = GgrsConfig<BoxInput>;
```

I show a type definition here, as I suspect end users would prefer this option for the sake of brevity. Because this is all done through the addition of a simple zero-sized-type, all existing applications can continue using their custom `Config` types if they want, with no breaking changes.

### DIFF
--- a/examples/box_game/box_game.rs
+++ b/examples/box_game/box_game.rs
@@ -1,10 +1,10 @@
 use bevy::{prelude::*, utils::HashMap};
 use bevy_ggrs::{
-    AddRollbackCommandExtension, LocalInputs, LocalPlayers, PlayerInputs, Rollback, Session,
+    AddRollbackCommandExtension, GgrsConfig, LocalInputs, LocalPlayers, PlayerInputs, Rollback,
+    Session,
 };
 use bytemuck::{Pod, Zeroable};
-use ggrs::Config;
-use std::{hash::Hash, net::SocketAddr};
+use std::hash::Hash;
 
 const BLUE: Color = Color::rgb(0.8, 0.6, 0.2);
 const ORANGE: Color = Color::rgb(0., 0.35, 0.8);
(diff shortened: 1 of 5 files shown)


johanhelsing (approved):
I like this. For the extreme_bevy tutorial, it means:

```diff
-struct GgrsConfig;
-
-impl ggrs::Config for GgrsConfig {
-    // 4-directions + fire fits easily in a single byte
-    type Input = u8;
-    type State = u8;
-    // Matchbox' WebRtcSocket addresses are called `PeerId`s
-    type Address = PeerId;
-}
+type ExtremeConfig = GgrsConfig<u8, PeerId>;
```

gschup (approved):
This is great syntactic sugar :)

Issue #71, opened 2023-10-16
Desync detection appears to compare the checksum of predicted frames to confirmed frames


nezuo:
**Describe the bug**
Desyncs are reported because predicated checksums and confirmed checksums are being compared.

**To Reproduce**
My repro can be found here on the `desync-detection-issue` branch: <https://github.com/gschup/bevy_ggrs/compare/main...nezuo:bevy_ggrs:desync-detection-issue?expand=1>

I used clumsy so that it predicts inputs/rollbacks.

**Expected behavior**
Only the checksums of confirmed frames should be compared.

**Screenshots**
Frame 18s for player 1 (in order):
![image](https://github.com/gschup/bevy_ggrs/assets/48431591/ae2735c0-b100-4f08-b9d4-060d34c2c60d)
![image](https://github.com/gschup/bevy_ggrs/assets/48431591/17bac232-1657-461b-aeed-bbfe353390d5)

Frame 18 for player 2:
![image](https://github.com/gschup/bevy_ggrs/assets/48431591/eba05192-68fe-4626-9fde-4437058f6027)

The second frame 18 for player 1 is the same as player 2's frame 18. However, it's comparing the first frame 18 with player 2's frame 18.

PraxTube:
Yeah, checksums as they are right now only work when there is no ping
(or the ping is lower then the input delay), which is almost only the case
when testing locally.

However when testing with high ping you get desyncs non-stop
because the receiving client is constantly mispredicting and we have to rollback a bunch.

Comparing the checksum of only confirmed frames would fix this?

johanhelsing:
Thanks for taking the time looking into this! It does indeed not happen if input latency is higher than ping.

I didn't catch this earlier, since my game breaks on rollbacks due to the regression #63

johanhelsing:
This is bug in `ggrs`. I created a bug and a fix for it there (see referenced issues above)

gschup:
The fix is now merged in the ggrs repo, so the issue can be closed here, as well! (unless I am missing something)

johanhelsing:
Just wondering if it's good to keep it open until at least the main branch of this repo depends on a ggrs version that works (just for visibility in case this is tripping people up)

gschup:
You are right. I was under the impression that the main branch of this repo depends on the main branch of ggrs. At least that was my original intention in the past. It must have slipped through my fingers that we don't do that currently.

gschup:
[...]

PR #74, opened 2023-10-19
Add a Baseline Mapping to Preserve Relationships

### LINKED ISSUE #63: Adding non-rollbacked children to rollbacked entities break the hierarchy
**Describe the bug**

I'd like it to be possible to spawn cosmetic entities as children of rolled back entities.

Used to work prior to https://github.com/gschup/bevy_ggrs/pull/31

For instance to:

- compose a character of several sprites
- use 3rd-party plugins for particle effects

```rust
pub fn spawn_character_particles(
    commands: &mut Commands,
    character_entity: Entity,
    sprites: &Res<SpriteAssets>,
) {
    info!("Adding particles to character: {character_entity:?}");
    commands
        .entity(character_entity)
        .with_children(|character| {
            character
                .spawn((
                    Name::new("JetPackParticles"),
                    ParticleSystemBundle {
                        // transform: Transform::from_xyz()
                        particle_system: ParticleSystem {
                            max_particles: 10_000,
                            texture: ParticleTexture::Sprite(sprites.particle.clone()),
                            spawn_rate_per_second: 40.0.into(),
                            emitter_shape: CircleSegment {
                                radius: 0.0.into(),
                                opening_angle: 0.5,
                                direction_angle: PI * 3. / 2.,
                            }
                            .into(),

                            initial_speed: JitteredValue::jittered(1.0, -0.5..0.5),
                            // acceleration: todo!(),
                            lifetime: JitteredValue::jittered(0.3, -0.1..0.1),
                            color: ColorOverTime::Gradient(Curve::new(vec![
                                CurvePoint::new(Color::WHITE, 0.0),
                                CurvePoint::new(Color::NONE, 1.0),
                            ])),
                            scale: ValueOverTime::Constant(0.0005),
                            looping: true,
                            system_duration_seconds: 10.0,
                            z_value_override: Some(JitteredValue::new(55.0)),
                            // bursts: todo!(),
                            // space: todo!(),
                            ..default()
                        },
                        ..default()
                    },
                    bevy_particle_systems::Playing,
                    JetPackParticles,
                ))
                .log_entity_id("spawned jetpack_particles");
        });
}
```

Currently, this results in `Children` containing "dead" entities after the first time a snapshot restore happens.

**To Reproduce**
1. spawn an entity with `.add_rollback()`, and a child without `.add_rollback()`
2. trigger a rollback
3. traverse `Children` hierarchy, it now points to a "dead" entity

**Expected behavior**

I would expect the entity ids of the children without `Rollback` components to be left alone (like they used to).

**Describe alternatives you've considered**

Adding "cosmetic follower entities" in a separate hierarchy branch that contain a reference to the rolled back entity, and have a book-keeping step where entities with dangling refs are despawned, and positions are synced.

**Additional context**

The problem is that `map_entities` for unknown entities returns "dead" entities: https://github.com/bevyengine/bevy/blob/8ace2ff9e361dd7ef1bc620b84674def0cb56454/crates/bevy_hierarchy/src/components/children.rs#L26

https://github.com/bevyengine/bevy/issues/6790

bushrat011899:
# Objective

- Fixes #63

## Solution 1 (original)

Before populating the `EntityMap` used during rollback with tracked entities, all entities within the `World` are added as no-op mappings:

```rust
for entity in world.iter_entities() {
    let entity = entity.id();
    entity_map.insert(entity, entity); // no-op mapping
}
```

This does add additional overhead, so it may be worth pursuing alternate strategies, perhaps through a PR to Bevy itself.

## Solution 2 (current; suggestion from @johanhelsing)

Reverse the mappings for any dead entities, and apply the map again:

```rust
// At this point, the map should only contain rollback entities
debug_assert_eq!(entity_map.len(), rollback_entities.len());

// For every type that reflects `MapEntities`, map the entities so that they reference the
// new IDs after applying the snapshot.
for reflect_map_entities in type_registry
    .iter()
    .filter_map(|reg| reg.data::<ReflectMapEntities>())
{
    reflect_map_entities.map_all_entities(world, &mut entity_map)
}

// If the entity map is now larger than the set of rollback entities, then dead entities were created
if entity_map.len() > rollback_entities.len() {
    // Reverse dead-mappings, no-op correct mappings
    for original_entity in entity_map.keys().collect::<Vec<_>>() {
        let mapped_entity = entity_map.remove(original_entity).unwrap();

        if rollback_entities.remove(&original_entity) {
            // Rollback entity was correctly mapped; no-op
            entity_map.insert(mapped_entity, mapped_entity);
        } else {
            // An untracked bystander was mapped to a dead end; reverse
            entity_map.insert(mapped_entity, original_entity);
        }
    }

    // Map entities a second time, fixing dead entities
    for reflect_map_entities in type_registry
        .iter()
        .filter_map(|reg| reg.data::<ReflectMapEntities>())
    {
        reflect_map_entities.map_all_entities(world, &mut entity_map)
    }
}
```

This avoids the allocation of an `EntityMap` the size of all entities in the `World`, but does require running `map_all_entities` twice.

### DIFF
--- a/src/world_snapshot.rs
+++ b/src/world_snapshot.rs
@@ -140,6 +140,12 @@ impl WorldSnapshot {
         // Mapping of the old entity ids ( when snapshot was taken ) to new entity ids
         let mut entity_map = EntityMap::default();
 
+        // Include a no-op mapping as a baseline to preserve relationships between rollback and non-rollback entities
+        for entity in world.iter_entities() {
+            let entity = entity.id();
+            entity_map.insert(entity, entity);
+        }
+
         // first, we write all entities
         for rollback_entity in self.entities.iter() {
             // find the corresponding current entity or create new entity, if it doesn't exist
--- a/src/world_snapshot.rs
+++ b/src/world_snapshot.rs
@@ -253,6 +253,9 @@ impl WorldSnapshot {
             }
         }
 
+        // At this point, the map should only contain rollback entities
+        debug_assert_eq!(entity_map.len(), rollback_entities.len());
+
         // For every type that reflects `MapEntities`, map the entities so that they reference the
         // new IDs after applying the snapshot.
         for reflect_map_entities in type_registry
@@ -262,25 +265,28 @@ impl WorldSnapshot {
             reflect_map_entities.map_all_entities(world, &mut entity_map)
         }
 
-        // Reverse dead-mappings, no-op correct mappings
-        for original_entity in entity_map.keys().collect::<Vec<_>>() {
-            let mapped_entity = entity_map.remove(original_entity).unwrap();
+        // If the entity map is now larger than the set of rollback entities, then dead entities were created
+        if entity_map.len() > rollback_entities.len() {
+            // Reverse dead-mappings, no-op correct mappings
+            for original_entity in entity_map.keys().collect::<Vec<_>>() {
+                let mapped_entity = entity_map.remove(original_entity).unwrap();
 
-            if rollback_entities.remove(&original_entity) {
-                // Rollback entity was correctly mapped; no-op
-                entity_map.insert(mapped_entity, mapped_entity);
-            } else {
-                // An untracked bystander was mapped to a dead end; reverse
-                entity_map.insert(mapped_entity, original_entity);
+                if rollback_entities.remove(&original_entity) {
+                    // Rollback entity was correctly mapped; no-op
+                    entity_map.insert(mapped_entity, mapped_entity);
+                } else {
+                    // An untracked bystander was mapped to a dead end; reverse
+                    entity_map.insert(mapped_entity, original_entity);
+                }
             }
-        }
 
-        // Map entities a second time, fixing dead entities
-        for reflect_map_entities in type_registry
-            .iter()
-            .filter_map(|reg| reg.data::<ReflectMapEntities>())
-        {
-            reflect_map_entities.map_all_entities(world, &mut entity_map)
+            // Map entities a second time, fixing dead entities
+            for reflect_map_entities in type_registry
+                .iter()
+                .filter_map(|reg| reg.data::<ReflectMapEntities>())
+            {
+                reflect_map_entities.map_all_entities(world, &mut entity_map)
+            }
         }
     }
 }
--- a/src/world_snapshot.rs
+++ b/src/world_snapshot.rs
@@ -2,7 +2,7 @@ use bevy::{
     ecs::{entity::EntityMap, reflect::ReflectMapEntities},
     prelude::*,
     reflect::{Reflect, TypeRegistry},
-    utils::HashMap,
+    utils::{HashMap, HashSet},
 };
 use std::{fmt::Debug, num::Wrapping};
 
@@ -137,14 +137,11 @@ impl WorldSnapshot {
         let type_registry = type_registry.read();
         let mut rid_map = rollback_id_map(world);
 
+        // Keep track of entities which are being rolled back
+        let mut rollback_entities = HashSet::new();
+
         // Mapping of the old entity ids ( when snapshot was taken ) to new entity ids
         let mut entity_map = EntityMap::default();
-
-        // Include a no-op mapping as a baseline to preserve relationships between rollback and non-rollback entities
-        for entity in world.iter_entities() {
-            let entity = entity.id();
-            entity_map.insert(entity, entity);
-        }
 
         // first, we write all entities
         for rollback_entity in self.entities.iter() {
@@ -155,6 +152,9 @@ impl WorldSnapshot {
 
             // Add the mapping from the old entity ID to the new entity ID
             entity_map.insert(rollback_entity.entity, entity);
+
+            // Keep track of rollback entities for later...
+            rollback_entities.insert(rollback_entity.entity);
 
             // for each registered type, check what we need to do
             for registration in type_registry.iter() {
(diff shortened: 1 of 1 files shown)


johanhelsing:
I was considering iterating over `Query<Children, With<Rollback>>`, and adding only those... It would solve the specific problem of hierarchies, but not the more fundamental general problem.

bushrat011899:
> I was considering iterating over `Query<Children, With<Rollback>>`, and adding only those... It would solve the specific problem of hierarchies, but not the more fundamental general problem.

Agreed. I think it makes more sense to get a working but possibly slow solution now, instead of trying to get a partially correct but faster solution later. Premature optimization is the root of all evil, etc.

johanhelsing:
Okay, I think we have the following options:

1. Add identity mappings for all entities (this pr)
2. Add identity mappings only for child/parent relationships, leave other relationships broken
3. Revert #31 (which introduced the hierarchy-destroying regression), do a holistic non-regressive change for adding relationship support.
4. Try to get API into bevy for a version of entity mapping that preserves unknown entities.

There are also rumors that Bevy will fundamentally change how they do relationships.

Currently, I'm leaning towards 1, possibly with a way to opt out, and just wait to see what bevy comes up with.

Curious to hear if @gschup has any thoughts on this.

johanhelsing:
Had an idea for a possible horrible hack:

1. Run entity mapping as we currently do
2. Create a new entity map
3. Iterate through old entity map, add reverse mappings for any dead entries in the map. Add identity mappings for all others.
4. Run entity mapping again with the second entity map

bushrat011899:
> Had an idea for a possible horrible hack:
> 
> 1. Run entity mapping as we currently do
> 2. Create a new entity map
> 3. Iterate through old entity map (that has now gotten entries for the dead entities), add reverse mappings for any dead entries in the old map to the new map. Add identity mappings for all others.
> 4. Run entity mapping again with the second entity map

Interesting idea! I'll give that a try and see how it goes.

bushrat011899:
Ok so I've pushed up an alteration to this PR which implements the concept @johanhelsing proposed. Instead of creating one large `EntityMap` containing all entities, and then updating the entries tracked for rollback, this now:

1. Creates the original `EntityMap` that only contains rollback entities, and a `HashSet` of the rollback entities for later use.
2. Apply the map.
3. Goes through each entry in the map and either replaces it with a no-op (`A -> B` becomes `B -> B`) if it was a rollback entity, or reverses the mapping if it was not (`A -> B` becomes `B -> A`). The reversal will only apply to dead entities, since they are the only type that could be in the map, but not in the set.
4. Apply the map a second time.

It is possible to avoid using the `HashSet` by checking if an entity is dead (they are supposed to have a specific signature to their inner values, making them identifiable), however I avoided adding that complexity because I don't like the idea of investigating the index and generation of an entity to determine if it's dead, since we can easily construct a list of known-living entities instead.

It seems to work, but I'd like @johanhelsing to test it with a project that was previously broken to confirm it actually resolves the issue.

Also, I don't know if it's faster than just making a big entity map at the beginning. I couldn't see an obvious problem in the `box_example` `sync_test`, but that's hardly demanding.

gschup:
On first glance, this seems like a good initial working solution. I wonder if there is a smarter way that avoids going through the map twice. In any case, we can figure optimizing this once it is no longer broken.

bushrat011899:
> On first glance, this seems like a good initial working solution. I wonder if there is a smarter way that avoids going through the map twice. In any case, we can figure optimizing this once it is no longer broken.

As you were typing this I noticed that we can get a slight optimisation by just checking if the entity map grew after applying it the first time. If it didn't, then no dead entities were created, so a second mapping is not required.

src/world_snapshot.rs
```diff
@@ -253,11 +253,39 @@ impl WorldSnapshot {
...
-            if let Some(map_entities_reflect) = registration.data::<ReflectMapEntities>() {
-                map_entities_reflect.map_all_entities(world, &mut entity_map)
+        for reflect_map_entities in type_registry
+            .iter()
+            .filter_map(|reg| reg.data::<ReflectMapEntities>())
+        {
+            reflect_map_entities.map_all_entities(world, &mut entity_map)
+        }
+
+        // If the entity map is now larger than the set of rollback entities, then dead entities were created
```
johanhelsing, line 268: Would it perhaps be good to explain in a comment here somewhere why this workaround is needed? i.e. that we can't preserve unknown entity ids with the provided bevy api.

johanhelsing:
> It seems to work, but I'd like @johanhelsing to test it with a project that was previously broken to confirm it actually resolves the issue.

I'll give it a go. There might be other rollback issues in my game, though, bevy 0.10 -> 0.11 has been a tough one for me.

Code looks sensible to me. I agree this is a good baseline. Should be correct, and adds relatively minimal (constant time) overhead. Not a bad deal to fix a serious regression.

johanhelsing:
Seems to be working like a charm. No more broken parent-child relationships in my game 🎉 

My game is still a bit broken due to #71, I a bunch of false positive (i think at least) desync errors so not a perfect test case yet. But still think it should be good to merge this once the conflicts are resolved 👍

johanhelsing:
Hmm... There is a quite noticeable performance drop in synctest sessions, though (in debug mode at least)

johanhelsing:
> Hmm... There is a quite noticeable performance drop in synctest sessions, though (in debug mode at least)

lol. scratch that. I'm now pretty sure it was just my hierarchy sanity checking system that was running a lot more efficiently when hierarchies were actually broken

gschup (approved):
For now, I like the solution. The code in world_snapshot.rs is very monolithic, but that is something we can maybe approach once it works properly! Thanks for doing the work!

bushrat011899:
> For now, I like the solution. The code in world_snapshot.rs is very monolithic, but that is something we can maybe approach once it works properly! Thanks for doing the work!

No problems! I look forward to helping out more on the future 😁

mikeder:
FWIW I found that parent->child transform propagation is still different/broken compared to before the bevy `0.11` updates.

This does not work:
```bash
.
└── parent
    └── child1 <- follows parent just fine
        └── child2 <- does not follow either parent unless with .add_rollback()
```

This does work:
```bash
.
└── parent
    ├── child1
    └── child2
```

johanhelsing:
bevy_ggrs does not do anything to propagate transforms, i don't think it did before, and I'm not sure it should either...

This or was about the entity *relations* (what entity have what children) which I think should be ok now.

mikeder:
@johanhelsing understood, just commenting as I found my original implementation broken by the Bevy 0.11 upgrade and thought it to be related. It is curious that using `.add_rollback()` in the first example above resulted in the desired behavior again, despite not doing anything to propagate the transforms.

johanhelsing:
Gotcha. If you think something is still broken wrt hierarchies, it's probably best to open a new issue.

PR #75, opened 2023-10-19
Removed Redundant `RwLock` and `Arc`; Removed `parking_lot` Dependency

bushrat011899:
# Objective

The `RollbackTypeRegistry` contains a `TypeRegistry`, which itself is a thin wrapper around `TypeRegistryInternal`, but through an `Arc<RwLock<...>>`. This indirection adds overhead to every interaction with the rollback type registry for no particular reason and should be removed. As a bonus, removing it would remove the need for the `parking_lot` dependency (still in the graph because of Bevy)

## Solution

Stored `TypeRegistryInternal` inside the `RollbackTypeRegistry` instead and removed lock acquisition code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ instant = { version = "0.1", optional = true }
 log = "0.4"
 ggrs = { version= "0.9.4", features=["sync-send"]}
 #ggrs = { git = "https://github.com/gschup/ggrs", features=["sync-send"]}
-parking_lot = "0.12.1"
 
 [dev-dependencies]
 structopt = "0.3"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,12 +4,11 @@
 use bevy::{
     ecs::schedule::{LogLevel, ScheduleBuildSettings, ScheduleLabel},
     prelude::*,
-    reflect::{FromType, GetTypeRegistration, TypeRegistry, TypeRegistryInternal},
+    reflect::{FromType, GetTypeRegistration, TypeRegistryInternal},
     utils::{Duration, HashMap},
 };
 use ggrs::{Config, InputStatus, P2PSession, PlayerHandle, SpectatorSession, SyncTestSession};
-use parking_lot::RwLock;
-use std::{marker::PhantomData, sync::Arc};
+use std::marker::PhantomData;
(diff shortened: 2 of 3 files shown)


johanhelsing:
Is this a performance problem in practice (root of all-evil and all that)?

Doesn't the `Internal` suffix mean it's considered private Bevy API?

bushrat011899:
> Is this a performance problem in practice (root of all-evil and all that)?

To my knowledge, no, and I'd be surprised if any benchmark could even reliably measure any difference. My goal is just to simplify that part, since the `Arc` and `RwLock` don't provide any benefits in this case.

> Doesn't the `Internal` suffix mean it's considered private Bevy API?

Not in this case. The reason for `TypeRegistry` and `TypeRegistryInternal` is that because of how scenes and assets are loaded, the `TypeRegistry` needs to be accessed in a way that's only possible through `Arc<RwLock<...>>`. For `bevy_ggrs`, that requirement does not exist. If you look at the definition for `TypeRegistry` in Bevy, you'll see a comment explaining the eventual desire to remove it entirely in favour of `TypeRegistryInternal`. You can even see it through the type names, as `TypeRegistryInternal` is actually named `TypeRegistry`, while `TypeRegistry` is named `TypeRegistryArc`.

gschup (approved):
I see why using the TypeWriterInternal is better here, good catch. When I wrote this, I think I just took the TypeWriter usage from some bevy code. 

Since bevy is going to change a lot anyway, I think it is fine to use this as is.

johanhelsing (approved):
> Not in this case. The reason for `TypeRegistry` and `TypeRegistryInternal` is that because of how scenes and assets are loaded, the `TypeRegistry` needs to be accessed in a way that's only possible through `Arc<RwLock<...>>`. For `bevy_ggrs`, that requirement does not exist. If you look at the definition for `TypeRegistry` in Bevy, you'll see a comment explaining the eventual desire to remove it entirely in favour of `TypeRegistryInternal`. You can even see it through the type names, as `TypeRegistryInternal` is actually named `TypeRegistry`, while `TypeRegistry` is named `TypeRegistryArc`.

Excellent. In that case, I have no objections. And nice with simpler code :)

PR #76, opened 2023-10-19
Refactor into Modular Plugins

bushrat011899:
# Objective

For a [long time now](https://github.com/gschup/bevy_ggrs/pull/52#issuecomment-1722489533), there's been a desire to avoid the requirement for `Reflection` in order to rollback components and resources. Since #66 has been merged, we now have the option of adding completely custom rollback behaviour either as an end user, or within `bevy_ggrs`.

## Solution

I have created a selection of plugins, which (with some helper data structures) demonstrate how we could add completely custom rollback behaviour. Because these plugins don't overlap on any mutable parameters (with some reasonable exceptions), their systems can run entirely in parallel, as scheduled by Bevy. This has the potential to provide a massive performance improvement, even without smarter implementations. For example, the Bevy smart points (`Res`, `ResMut`, `Mut`, `Ref`, etc.) all contain a `Ticks` item, which records when items were last added/changed. It would be pretty trivial to store that information along side the snapshot, avoiding rollback entirely for unchanged items, and thus avoiding noisy change detection (`Changed`, `Added`, etc.) even without `bypass_change_detection`.

## Usage

When setting up an app, end users can either use extension methods provided by `GgrsApp`:

```rust
app.rollback_resource_with_clone::<FrameCount>()
    .rollback_component_with_clone::<Transform>()
    .rollback_component_with_clone::<Velocity>();
```

Or by adding the underlying plugins directly:

```rust
app.add_plugins((
    ResourceSnapshotPlugin::<CloneStrategy<FrameCount>>::default(),
    ComponentSnapshotPlugin::<CloneStrategy<Transform>>::default(),
    ComponentSnapshotPlugin::<CloneStrategy<Velocity>>::default()
));
```

#### `GgrsPlugin`

The main entry point for `bevy_ggrs`, providing a sensible default starting point.

```rust
app.add_plugins(GgrsPlugin::<MyGameConfig>::default());
```

#### `EntitySnapshotPlugin`

A [`Plugin`] which manages the rollback for [`Entities`](`Entity`). This will ensure all [`Entities`](`Entity`) match the state of the desired frame, or can be mapped using a [`RollbackEntityMap`], which this [`Plugin`] will also manage.

This can be considered one of the core plugins, since it is responsible for ensuring the entity graph itself is ready, allowing other systems to manage resources and components.

#### `ComponentSnapshotPlugin<S>`

A [`Plugin`] which manages snapshots for a [`Component`] a provided [`Strategy`] `S`.

```rust
#[derive(Component, Clone)]
pub struct MyVectorOfData(/* ... */);

app.add_plugins(ComponentSnapshotPlugin::<CloneStrategy<MyVectorOfData>>::default());
```

The three built-in strategies provided are:

- `CloneStrategy`: Clones data and stores as the original type.
- `CopyStrategy`: Copies data and stores as the original type.
- `ReflectStrategy`: Reflects the data, clones it, and stores as `Box<dyn Reflect>`.

#### `ResourceSnapshotPlugin<S>`

A [`Plugin`] which manages snapshots for a [`Resource`] a provided [`Strategy`] `S`.

```rust
#[derive(Resource, Clone)]
pub struct MyVectorOfData(/* ... */);

app.add_plugins(ResourceSnapshotPlugin::<CloneStrategy<MyVectorOfData>>::default());
```

The three built-in strategies provided are:

- `CloneStrategy`: Clones data and stores as the original type.
- `CopyStrategy`: Copies data and stores as the original type.
- `ReflectStrategy`: Reflects the data, clones it, and stores as `Box<dyn Reflect>`.

#### `ChecksumPlugin`

A [`Plugin`] which creates a [`Checksum`] resource which can be read after the [`SaveWorldSet::Snapshot`] set in the [`SaveWorld`] schedule has been run. To add you own data to this [`Checksum`], create an [`Entity`] with a [`ChecksumPart`] [`Component`]. Every [`Entity`] with this [`Component`] will participate in the creation of a [`Checksum`].

```rust
// This checksum is updated every frame in SaveWorld
let checksum = world.get_resource::<Checksum>();
```

Note that this plugin does not store [`Checksums`] for older frames.

#### `ComponentChecksumHashPlugin<C>`

A [`Plugin`] which will track the [`Component`] `C` on [`Rollback Entities`](`Rollback`) and ensure a [`ChecksumPart`] is available and updated. This can be used to generate a [`Checksum`](`crate::Checksum`).

```rust
fn get_checksum_for_foo(parts: Query<&ChecksumPart, With<<ChecksumFlag<Foo>>>>) {
    let mut hasher = DefaultHasher::new();

    for part in parts.iter() {
        part.hash(&mut hasher);
    }

    let checksum: u64 = hasher.finish();
}
```

#### `ComponentMapEntitiesPlugin<C>`

A [`Plugin`] which updates the state of a post-rollback [`Component`] `C` using [`MapEntities`].

```rust
#[derive(Component, Clone, Copy)]
pub struct BestFriend(Entity);

impl MapEntities for BestFriend {
    /* ... */
}

app.add_plugins((
    ComponentSnapshotCopyPlugin::<BestFriend>::default(),
    ComponentMapEntitiesPlugin::<BestFriend>::default(),
));
```

#### `SnapshotSetPlugin`

Sets up the [`LoadWorldSet`] and [`SaveWorldSet`] sets, allowing for explicit ordering of rollback systems across plugins.

```rust
app.add_systems(LoadWorld, log_pre_mapping_parents.before(LoadWorldSet::Mapping));
```

#### `ResourceChecksumHashPlugin<R>`

Plugin which will track the [`Resource`] `R` and ensure a [`ChecksumPart`] is available and updated. This can be used to generate a [`Checksum`](`crate::Checksum`).

```rust
fn get_checksum_for_bar(parts: Query<&ChecksumPart, With<<ChecksumFlag<Bar>>>>) {
    let mut hasher = DefaultHasher::new();

    for part in parts.iter() {
        part.hash(&mut hasher);
    }

    let checksum: u64 = hasher.finish();
}
```

#### `ResourceMapEntitiesPlugin<R>`

A [`Plugin`] which updates the state of a post-rollback [`Resource`] `R` using [`MapEntities`].

```rust
#[derive(Resource, Clone, Copy)]
pub struct Player(Entity);

impl MapEntities for Player {
    /* ... */
}

app.add_plugins((
    ResourceMapEntitiesPlugin::<Player>::default(),
    ResourceMapEntitiesPlugin::<Player>::default(),
));
```

## Outstanding Issues

- [X] ~~Components and resources which require mapping don't have an easy way to gain access to the `EntityMap` generated during `load_world`. I think this should be added as a resource to the `LoadWorld` schedule as a simple fix.~~ Got a map resource available now, ~~with some bugs~~. Bugs are now gone!
- [X] I'm currently storing snapshots with a fixed depth of 60. This depth is configurable, ~~but it should be set automatically based on the session (I believe GGRS provides a "max rollback distance" parameter we could use)~~. I'm now asking the `Session` for the current `confirmed_frame`, and anything older than that is discarded.
- [X] ~~I've simply added these plugins and not integrated them with `GgrsPlugin` or `GgrsApp`. I think it might be desirable to have a clean way to add rollback functionality like `register_rollback_component`.~~ All plugins are accessible through `GgrsApp` methods.
- [X] ~~I believe it is possible to create a more generic `RollbackResourcePlugin`/`RollbackComponentPlugin` that can accept a `Strategy`, which would be an object describing how to store, load, and update a rollback item.~~ Implemented as described.


johanhelsing:
Thanks for working one this!

I really like the modularity of this approach. Feels elegant that you could just write your own plugin if the ones we provide aren't a good fit. Relatively easy to understand how it works.

> Components and resources which require mapping don't have an easy way to gain access to the EntityMap generated during load_world. I think this should be added as a resource to the LoadWorld schedule as a simple fix.

Yes, would probably make sense to expose a (sane) entity map as a resource... Would probably need to be two steps? One for adding mappings, and one for applying them... Not sure how that would work with our hack?

> I'm currently storing snapshots with a fixed depth of 60. This depth is configurable, but it should be set automatically based on the session (I believe GGRS provides a "max rollback distance" parameter we could use)

Maybe another approach is to provide access to the latest confirmed frame, and just clear frames older than that?

bushrat011899:
Ok so this is turning into a pretty major re-factor of `bevy_ggrs`, but I'm in too deep to turn around now. I've completely removed the `world_snapshot.rs` file and everything it contained. Its responsibilities have instead been compartmentalised into dedicated plugins with narrowed systems. Now, instead of a pair of large `save_world` and `load_world` systems that have exclusive access to the world, there are the following plugins:

- `GgrsChecksumPlugin`: Responsible for taking **reported** checksums and combining them into a single value just for the current frame.
- `GgrsComponentChecksumHashPlugin<C>`: Responsible for computing the checksum of one component type `C`, using `Hash`.
- `GgrsComponentSnapshotClonePlugin<C>`: Responsible for saving/loading the component `C` using `Clone`.
- `GgrsComponentMapEntitiesPlugin<C>`: Responsible for updating the entity relationships for the component `C` using `MapEntities`.
- `GgrsComponentSnapshotReflectPlugin<C>`: Responsible for saving/loading the component `C` using `Reflect + FromWorld`.
- `GgrsEntitySnapshotPlugin`: Responsible for snaphottin and de/spawning entities during a rollback, and also creates a `RollbackEntityMap` that can be used by other systems.
- `GgrsResourceSnapshotClonePlugin<R>`: Responsible for saving/loading the resource `R` using `Clone`.
- `GgrsResourceMapEntitiesPlugin<R>`: Responsible for updating the entity relationships for the resource `R` using `MapEntities`.
- `GgrsResourceSnapshotReflectPlugin<R>`: Responsible for saving/loading the component `C` using `Reflect + FromWorld`.

Using these plugins, the `TypeRegistry` is completely redundant and has been removed. To get back to pre-PR behaviour, the `GgrsPlugin` has been modified to add the following default plugins:

```rust
app.add_plugins((
    GgrsChecksumPlugin,
    GgrsResourceSnapshotClonePlugin::<Checksum>::default(),
    GgrsEntitySnapshotPlugin,
    GgrsComponentSnapshotReflectPlugin::<Parent>::default(),
    GgrsComponentMapEntitiesPlugin::<Parent>::default(),
    GgrsComponentSnapshotReflectPlugin::<Children>::default(),
    GgrsComponentMapEntitiesPlugin::<Children>::default(),
));
```

While this is a lot of plugins (and I've placed each into their own file which creates a lot of verbosity compared to `world_snapshot.rs`), each is very clear and maintainable. For example, the plugin for getting component checksums is:

```rust
impl<C> GgrsComponentChecksumHashPlugin<C>
where
    C: Component + Hash,
{
    pub fn update(
        mut commands: Commands,
        components: Query<&C, (With<Rollback>, Without<ChecksumFlag<C>>)>,
        mut checksum: Query<&mut ChecksumPart, (Without<Rollback>, With<ChecksumFlag<C>>)>,
    ) {
        let mut hasher = DefaultHasher::new();

        let Ok(mut checksum) = checksum.get_single_mut() else {
            commands.spawn((ChecksumPart::default(), ChecksumFlag::<C>::default()));

            return;
        };

        for component in components.iter() {
            component.hash(&mut hasher);
        }

        *checksum = ChecksumPart(hasher.finish());
    }
}

impl<C> Plugin for GgrsComponentChecksumHashPlugin<C>
where
    C: Component + Hash,
{
    fn build(&self, app: &mut App) {
        app.add_systems(SaveWorld, Self::update);
    }
}
```

Because of how they're segregated along `Component` and `Resource` types, almost all of them can run in parallel, with the exception of some obvious dependency (e.g., can't rollback `Parent` until all the entities have been de/spawned, can't map until the `Parent` is rolled back, etc.). I still have to add some system sets to organise these dependencies, and there's currently a bug with commands not being flushed at the right time for entity mapping, but the `box_game` `sync_test` runs with all these changes, and so does the P2P example.

johanhelsing:
Love it!

> Using these plugins, the `TypeRegistry` is completely redundant and has been removed.

I wonder if this affects performance?

> To get back to pre-PR behaviour, the `GgrsPlugin` has been modified to add the following default plugin

This seems super-clean to me :)

bushrat011899:
> Love it!

Thanks!

> > Using these plugins, the `TypeRegistry` is completely redundant and has been removed.
> 
> I wonder if this affects performance?

Honestly unsure. I think it would help with compiler optimisations, since the type information will be held onto all the way til it's taken by Bevy itself. What I am confident in saying is this framework would allow for the creation of the optimal solution. For example, it may be much better to create a custom plugin for adding/removing `Parent` and `Children` using the built-in Bevy commands. With this new framework, that is entirely possible, and can be done by an end user at their discretion.

> > To get back to pre-PR behaviour, the `GgrsPlugin` has been modified to add the following default plugin
> 
> This seems super-clean to me :)

Again, thanks!

bushrat011899:
I'm marking this PR as ready for review because at this point I'd really appreciate some more specific feedback on possible changes. Currently, I believe this PR "just works", and doesn't even require any updates to existing games, since it mostly reuses existing API. A key exception is rollback resources/components added with `app.register_rollback_component(...)` will no longer perform entity mapping by default. I don't know how common that kind of problem would be though, and can be easily worked around by adding `app.add_plugins(GgrsComponentMapEntitiesPlugin::<...>::default())`.

I've made lots of very opinionated changes, but I'm completely open to revisiting any of my choices if they're a poor fit for the project.

src/snapshot/checksum.rs
```diff
@@ -0,0 +1,59 @@
...
+/// [`Component`]. Every [`Entity`] with this [`Component`] will participate in the
+/// creation of a [`Checksum`].
+pub struct GgrsChecksumPlugin;
+
+impl GgrsChecksumPlugin {
+    /// A [`System`] responsible for updating [`Checksum`] based on [`ChecksumParts`](`ChecksumPart`).
+    pub fn update(mut checksum: ResMut<Checksum>, parts: Query<&ChecksumPart>) {
+        let mut hasher = DefaultHasher::new();
+
+        for part in parts.iter() {
```
johanhelsing, line 46: are we sure the order is stable here?

src/snapshot/component_checksum_hash.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+}
+
+impl<C> Plugin for GgrsComponentChecksumHashPlugin<C>
+where
+    C: Component + Hash,
+{
+    fn build(&self, app: &mut App) {
+        app.add_systems(SaveWorld, Self::update.in_set(SaveWorldSet::Snapshot));
```
johanhelsing, line 62: Would it make sense with a checksum set (honest question don't know the code well enough yet)?

src/snapshot/component_copy.rs
```diff
@@ -0,0 +1,84 @@
+use crate::{
+    GgrsComponentSnapshot, GgrsSnapshots, LoadWorld, LoadWorldSet, Rollback, RollbackFrameCount,
+    SaveWorld, SaveWorldSet,
+};
+use bevy::prelude::*;
+use std::marker::PhantomData;
+
+/// A [`Plugin`] which manages snapshots for a [`Component`] `C` using [`Copy`].
+pub struct GgrsComponentSnapshotCopyPlugin<C>
```
johanhelsing, line 9: What do we gain by having this one in addition to the clone-based one?

src/snapshot/component_checksum_hash.rs
```diff
@@ -0,0 +1,64 @@
...
+    _phantom: PhantomData<C>,
+}
+
+impl<C> Default for GgrsComponentChecksumHashPlugin<C>
+where
+    C: Component + Hash,
+{
+    fn default() -> Self {
+        Self {
+            _phantom: Default::default(),
```
johanhelsing, line 26: nit: could use just `default()` here. repeats

src/snapshot/component_clone.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// A [`Plugin`] which manages snapshots for a [`Component`] `C` using [`Clone`].
+pub struct GgrsComponentSnapshotClonePlugin<C>
+where
+    C: Component + Clone,
+{
+    _phantom: PhantomData<C>,
+}
+
+type Snapshots<C> = GgrsSnapshots<C, GgrsComponentSnapshot<C>>;
```
johanhelsing, line 16: I notice a bit of friction trying to understand this... perhaps I'm just tired... I also see that it's repeated a lot...

Perhaps it would make sense to move this type alias to the library crate, and expose it under a slightly more descriptive name?

As far as I understand it could just be `ComponentSnapshots<C>`?

src/snapshot/component_clone.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl<C> Plugin for GgrsComponentSnapshotClonePlugin<C>
+where
+    C: Component + Clone,
+{
+    fn build(&self, app: &mut App) {
+        app.init_resource::<Snapshots<C>>()
+            .add_systems(
+                SaveWorld,
+                (Snapshots::<C>::discard_old_snapshots, Self::save)
```
johanhelsing, line 78: I like how you only had to implement discard_old_snapshots once :)

src/snapshot/component_clone.rs
```diff
@@ -0,0 +1,84 @@
...
+            _phantom: Default::default(),
+        }
+    }
+}
+
+impl<C> GgrsComponentSnapshotClonePlugin<C>
+where
+    C: Component + Clone,
+{
+    pub fn save(
```
johanhelsing, line 33: Just a style thing, but I haven't often seen bevy systems as associated functions, is this a new thing. Is there an advantage or reason over just using a plain generic function?

johanhelsing (commented):
I like the general direction and API changes.

I really wanted to finish reviewing this today, but I noticed I got a bit tired half-way through and my comments were getting sloppier.

Not sure when I will have time to continue, so posting what I've got so far, even though there most likely are some stupid questions in there :P

What I've read so far is great, though :) Also love that you documented things fairly well.

johanhelsing:
> `app.register_rollback_component(...)` will no longer perform entity mapping by default. I don't know how common that kind of problem would be though, and can be easily worked around by adding `app.add_plugins(GgrsComponentMapEntitiesPlugin::<...>::default())`.

IMO it would be better to break backwards source-compatibility and just make what we think is the best API going forwards

src/snapshot/component_checksum_hash.rs
bushrat011899: I don't think so. The way I see it, the `SaveWorldSet::Snapshot` should be an entirely read-only set for anything affecting the world. It would be trivial to add an extra set for it, but I don't think it would make a semantic difference.

src/snapshot/component_copy.rs
bushrat011899: Clone explicitly calls whatever cloning procedure the type implements, whereas Copy does a bitwise copy of the underlying memory. In theory, the Copy snapshot should be the absolute fastest without some domain specific tricks.

src/snapshot/component_clone.rs
bushrat011899: These are the little things I was really happy writing!

src/snapshot/component_clone.rs
bushrat011899: I mostly just did it for tidiness, since I can reuse the `where` clause across the whole type. I don't think it makes any difference under the hood.

src/snapshot/component_checksum_hash.rs
```diff
@@ -0,0 +1,64 @@
...
+{
+    /// A [`System`] responsible for managing a [`ChecksumPart`] for the [`Component`] type `C`.
+    pub fn update(
+        mut commands: Commands,
+        components: Query<&C, (With<Rollback>, Without<ChecksumFlag<C>>)>,
+        mut checksum: Query<&mut ChecksumPart, (Without<Rollback>, With<ChecksumFlag<C>>)>,
+    ) {
+        let mut hasher = DefaultHasher::new();
+
+        for component in components.iter() {
```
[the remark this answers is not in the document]
bushrat011899: Agreed. I've added explicit sorting by `Rollback`

src/snapshot/checksum.rs
bushrat011899: Probably not! I've moved this to use `wrapping_add` for the combination (order independent) and then hash the result at the end.

src/snapshot/component_clone.rs
bushrat011899: Yeah I think I can clarify this line a lot better. I'm popping out for a bit but when I come back I'll move these around and name them better. The basic explanation is I just needed to type out a fairly lengthy generic type 3 times, so I just made a shortcut to avoid a typo (which I had made a couple times before!)

bushrat011899: I've added some explicit type declarations, `GgrsResourceSnapshots` and `GgrsComponentSnapshots` (with documentation) which I believe better communicate their purpose.

bushrat011899:
> IMO it would be better to break backwards source-compatibility and just make what we think is the best API going forwards

I do agree with this sentiment, especially considering Bevy itself is still breaking its API at a somewhat regular pace (not a complaint, there's amazing progress being made!). I would like to have a crack at making a little plugin builder to facilitate the syntax we've discussed previously:

```rust
app.add_plugins((
    Rollback::for_resource<Foo>().with_clone().with_mapping(),
    Rollback::for_component<Bar>().with_copy(),
));
```

src/snapshot/component_copy.rs
bushrat011899: Yeah according to the [Rust docs](https://doc.rust-lang.org/std/marker/trait.Copy.html#whats-the-difference-between-copy-and-clone), `Copy` will always do a bitwise `memcopy`, even if you have a `Clone` implementation that's completely different.

> Copies happen implicitly, for example as part of an assignment y = x. The behavior of Copy is not overloadable; it is always a simple bit-wise copy.

Which is why `Copy` is allowed to be implicit, since it's virtually free. I think the confusion might be coming from the requirement that anything which is `Copy` must also be `Clone`. Which you would think implies `Copy` uses `Clone`, but it actually doesn't.

bushrat011899: Having said that, for types which can be `Copy`, I think their `Clone` is (usually) derived to be `*self`, so potentially `Clone` is as fast as `Copy`, but it can be slower if the implementation isn't that.

src/snapshot/component_checksum_hash.rs
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+}
+
+impl<C> Plugin for GgrsComponentChecksumHashPlugin<C>
+where
+    C: Component + Hash,
+{
+    fn build(&self, app: &mut App) {
+        app.add_systems(SaveWorld, Self::update.in_set(SaveWorldSet::Snapshot));
```
johanhelsing, line 66: Should probably have an explicit ordering before `GgrsChecksumPlugin::update`, or we will get checksums for the last frame.

src/snapshot/checksum.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+/// Represents a checksum value for a specific type, flagged by [`ChecksumFlag`].
+#[derive(Component, Default, Hash)]
+pub struct ChecksumPart(pub u64);
+
+/// Represents a total checksum for a given frame.
+#[derive(Resource, Default, Clone, Copy)]
+pub struct Checksum(pub u64);
```
johanhelsing, line 31: `u128` is used by ggrs, so maybe use that here as well?

src/snapshot/checksum.rs
bushrat011899: Yeah completely true, it would probably make sense to be its own schedule so it can be run whenever required.

src/schedule_systems.rs
```diff
@@ -85,75 +82,90 @@ pub(crate) fn run_synctest<C: Config>(world: &mut World, mut sess: SyncTestSessi
...
+
         match request {
             GGRSRequest::SaveGameState { cell, frame } => {
                 debug!("saving snapshot for frame {frame}");
                 world.run_schedule(SaveWorld);
 
                 // look into resources and find the checksum
                 let snapshots = world
-                    .get_resource::<RollbackSnapshots>()
+                    .get_resource::<GgrsSnapshots<Checksum, Option<Checksum>>>()
```
johanhelsing, line 159: Better to just get the `Checksum` resource directly here?

src/lib.rs
```diff
@@ -182,30 +132,63 @@ impl<C: Config> Plugin for GgrsPlugin<C> {
...
             .init_resource::<FixedTimestepData>()
             .add_schedule(GgrsSchedule, schedule)
             .add_schedule(ReadInputs, Schedule::new())
             .add_systems(PreUpdate, schedule_systems::run_ggrs_schedules::<C>)
-            .add_systems(LoadWorld, load_world)
-            .add_systems(SaveWorld, save_world);
+            .add_plugins((
+                GgrsSnapshotSetPlugin,
+                GgrsChecksumPlugin,
+                GgrsResourceSnapshotCopyPlugin::<Checksum>::default(),
```
johanhelsing, line 145: There is no guarantee this will run after the checksumming plugins, so if `GgrsResourceSnapshotCopyPlugin::save` runs before `GgrsChecksumPlugin::update` (which it does for me at least), we'll end up saving an outdated checksum in `handle_requests`.

johanhelsing (changes requested):

src/snapshot/component_copy.rs
johanhelsing: Don't think I've ever seen different Copy and Clone implementations.

But yeah, just wanted to make sure I didn't miss anything. So the value of this extra plugin is basically that the caller knows that the operation will be relatively cheap(-ish).

src/lib.rs
bushrat011899: Good spot! I didn't notice that at all, you're right in that there should be an explicit `Checksum` set to ensure the right checksum is saved.

src/snapshot/component_checksum_hash.rs
```diff
@@ -0,0 +1,68 @@
...
+where
+    C: Component + Hash,
+{
+    /// A [`System`] responsible for managing a [`ChecksumPart`] for the [`Component`] type `C`.
+    pub fn update(
+        mut commands: Commands,
+        components: Query<(&Rollback, &C), (With<Rollback>, Without<ChecksumFlag<C>>)>,
+        mut checksum: Query<&mut ChecksumPart, (Without<Rollback>, With<ChecksumFlag<C>>)>,
+    ) {
+        let mut hasher = DefaultHasher::new();
```
johanhelsing, line 41: > ```
> /// The default [`Hasher`] used by [`RandomState`].
> ///
> /// The internal algorithm is not specified, and so it and its hashes should
> /// not be relied upon over releases.
```

Maybe it would be better/more appropriate to use `bevy_utils::FixedState.build_hasher`, which is what reflect_hash uses?

File: src/schedule_systems.rs (hunk 11 above)
Comment (bushrat011899) [reply src/schedule_systems.rs:159]: Totally right! I forgot that the `frame` value provided in `SaveGameState` is supposed to be the current frame anyway.

File: src/snapshot/checksum.rs (hunk 10 above)
Comment (bushrat011899) [reply src/snapshot/checksum.rs:31]: Makes sense, I was following the type returned by `Hash`, but we don't actually know that `Hash` is even a good fit for the checksum anyway.

File: src/snapshot/component_checksum_hash.rs (hunk 13 above)
Comment (bushrat011899) [reply src/snapshot/component_checksum_hash.rs:41]: Didn't even know about that! Good idea!

**Comment by johanhelsing** [comment]
Okay, I got to sign off for today, will be a couple of day until I can pick it up again. Seems we're getting close, though :)

**Comment by bushrat011899** [comment]
> Okay, I got to sign off for today, will be a couple of day until I can pick it up again. Seems we're getting close, though :)

Yeah of course, take your time! Thanks so much for all your help with this, it's a lot of changes to go through and I really appreciate your advice.

**Review by nezuo — COMMENTED** [review]
Nice PR!

I just quickly reviewed and have a few minor comments.

File: src/lib.rs [hunk 14]
```diff
@@ -182,30 +132,63 @@ impl<C: Config> Plugin for GgrsPlugin<C> {
...
     }
 }
 
 /// Extension trait to add the GGRS plugin idiomatically to Bevy Apps
 pub trait GgrsApp {
-    /// Registers a component type for saving and loading from the world.
-    fn register_rollback_component<Type>(&mut self) -> &mut Self
+    /// Registers a component type for saving and loading from the world. This
+    /// uses [`Copy`] based snapshots for rollback.
+    fn register_rollback_component_with_copy<Type>(&mut self) -> &mut Self
```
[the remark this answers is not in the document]
johanhelsing: Yeah, the register is mostly there as a leftover from when it "registered in the type registry"

src/schedule_systems.rs
```diff
@@ -61,8 +60,8 @@ pub(crate) fn run_ggrs_schedules<T: Config>(world: &mut World) {
                 time_data.accumulator = Duration::ZERO;
                 time_data.run_slow = false;
                 world.insert_resource(LocalPlayers::default());
-                world.insert_resource(RollbackSnapshots::default());
                 world.insert_resource(RollbackFrameCount(0));
+                world.insert_resource(RollbackFrameConfirmed(0));
```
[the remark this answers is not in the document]
bushrat011899: Agreed, 0 implies the initial frame is already confirmed, good spot!

src/lib.rs
```diff
@@ -90,8 +88,9 @@ impl Default for FixedTimestepData {
 #[derive(Resource, Debug, Default)]
 pub struct RollbackFrameCount(i32);
 
-#[derive(Resource)]
-struct RollbackTypeRegistry(TypeRegistryInternal);
+/// The most recently confirmed frame. Any information for frames stored before this point can be safely discarded.
+#[derive(Resource, Debug, Default)]
+pub struct RollbackFrameConfirmed(i32);
```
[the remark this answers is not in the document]
bushrat011899: It's mostly to match the naming of `RollbackFrameCount`. IMO, both would need to be changed to maintain consistency.

src/lib.rs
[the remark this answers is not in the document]
bushrat011899: Agreed, the names should be shortened.

src/lib.rs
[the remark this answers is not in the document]
nezuo: I think my name suggestions are consistent. They both end with `FrameCount` but the prefix describes their functionality. `Confirmed` meaning it's only the confirmed frame count. `Rollback` meaning the resource rollbacks to the current frame.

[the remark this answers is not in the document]
bushrat011899: Yeah that makes more sense, I guess I didn't see it that way before, good point!

src/snapshot/checksum.rs
```diff
@@ -0,0 +1,60 @@
...
+/// Represents a checksum value for a specific type, flagged by [`ChecksumFlag`].
+#[derive(Component, Default, Hash)]
+pub struct ChecksumPart(pub u128);
+
+/// Represents a total checksum for a given frame.
+#[derive(Resource, Default, Clone, Copy)]
+pub struct Checksum(pub u128);
+
+/// A [`Plugin`] which creates a [`Checksum`] resource which can be read after the
+/// [`SaveWorldSet::Snapshot`] set in the [`SaveWorld`] schedule has been run.
```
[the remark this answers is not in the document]
bushrat011899: Yes that is correct, this docstring was out-of-date since the last commit, good spot!

examples/box_game/box_game_p2p.rs
```diff
@@ -63,10 +63,16 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        .register_rollback_resource::<FrameCount>()
+        // Rollback behavior can be controlled through many modular plugins
+        .add_plugins((
+            // The FrameCount resource implements Copy
+            // We can use that to have minimal overhead rollback
+            GgrsResourceSnapshotCopyPlugin::<FrameCount>::default(),
+            // Transform and Velocity components only implement Clone,
+            // so instead we'll use that to snapshot and rollback with
+            GgrsComponentSnapshotClonePlugin::<Transform>::default(),
+            GgrsComponentSnapshotClonePlugin::<Velocity>::default(),
```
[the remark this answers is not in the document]
bushrat011899: Agreed, it also looks cleaner in the examples now that it's all set up.

src/snapshot/resource_map.rs
```diff
@@ -0,0 +1,85 @@
...
+            apply_rollback_map_to_resource_inner::<R>(world, map);
+        });
+    }
+}
+
+fn apply_rollback_map_to_resource_inner<R>(world: &mut World, map: Mut<RollbackEntityMap>)
+where
+    R: Resource + MapEntities,
+{
+    let mut applied_entity_map = map.get_map();
```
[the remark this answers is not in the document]
bushrat011899: The problem is every mapping operation requires a mutable reference to the entity map (to allow for creating dead entities). A way we could mitigate this problem would be to actually create the `RollbackHierarchy` plugin I mentioned previously, where parent-child rollback is done through commands rather than entity mapping and reflection.

Honestly, I think the change to entity mapping Bevy made in 0.11 made the feature quite hostile for every case that isn't explicitly moving from one world to another. Instead of providing a mutable reference to an entity map, they probably should've passed it a `Commands` to allow the mapper to create entities as required explicitly.

src/schedule_systems.rs
```diff
@@ -85,75 +82,105 @@ pub(crate) fn run_synctest<C: Config>(world: &mut World, mut sess: SyncTestSessi
...
+        };
+
+        if let Some(confirmed_frame) = confirmed_frame {
+            world.insert_resource(ConfirmedFrameCount(confirmed_frame));
+        }
+
+        if let Some(Session::P2P(s)) = world.get_resource::<Session<T>>() {
+            let confirmed_frame = s.confirmed_frame();
+            world.insert_resource(ConfirmedFrameCount(confirmed_frame));
+        }
```
nezuo, line 170: Sorry if you planned to fix this already but the confirmed frame is being set for p2p sessions twice now.

nezuo (changes requested):

src/schedule_systems.rs
bushrat011899: Nope I missed that, whoops! Thanks!

src/snapshot/component_checksum_hash.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn update(
+        mut commands: Commands,
+        components: Query<(&Rollback, &C), (With<Rollback>, Without<ChecksumFlag<C>>)>,
+        mut checksum: Query<&mut ChecksumPart, (Without<Rollback>, With<ChecksumFlag<C>>)>,
+    ) {
+        let mut hasher = bevy::utils::FixedState::default().build_hasher();
+
+        let mut components = components.iter().collect::<Vec<_>>();
+
+        components.sort_by_key(|(&rollback, _)| rollback);
```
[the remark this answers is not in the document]
bushrat011899: Ok I spent some time and I think I have a good solution to this problem. What I've done is added a resource, `RollbackOrdered`, which contains a sorted list of every `Rollback` component ever created. This list is sorted and updating during `AddRollbackCommand` execution, which should avoid any runtime penalty, since all rollback entities should be registered during times when it is reasonable to be spawning entities anyway.

Doing it here allows me to add the index of the `Rollback` flag to the hash for each value in a component hash, making it unique for that component-rollback pairing. Then I can combine all those unique values using XOR, which is commutative (order independent).

The result is a checksum which is order stable, resistant to transposition, and doesn't require sorting.

src/snapshot/resource_map.rs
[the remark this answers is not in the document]
johanhelsing: I see... Maybe EntityMapper should have been a trait instead?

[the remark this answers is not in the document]
bushrat011899: Agreed. For now, I think we should leave the cloning in, and once this is merged I'll see if I can get a PR merged into Bevy which gives us some kind of alternative to mutable `EntityMapper` which suits are use-case better.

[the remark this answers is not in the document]
bushrat011899: Or, we remove the cloning but all mapping operations need to happen sequentially. Personally, I think cloning would be more performant and ergonomic to use, but that's more of a hunch than anything else.

src/schedule_systems.rs
```diff
@@ -85,75 +82,100 @@ pub(crate) fn run_synctest<C: Config>(world: &mut World, mut sess: SyncTestSessi
...
     for request in requests {
+        let current_frame = world
+            .get_resource::<RollbackFrameCount>()
+            .map(|frame| frame.0)
+            .unwrap_or_default();
+
+        let confirmed_frame = match world.get_resource::<Session<T>>() {
+            Some(Session::P2P(s)) => Some(s.confirmed_frame()),
+            Some(Session::SyncTest(s)) => {
+                // TODO: `max_prediction` should be replaced with `depth`, but it is private.
```
johanhelsing, line 155: What is depth?

src/snapshot/checksum.rs
```diff
@@ -0,0 +1,59 @@
...
+/// To add you own data to this [`Checksum`], create an [`Entity`] with a [`ChecksumPart`]
+/// [`Component`]. Every [`Entity`] with this [`Component`] will participate in the
+/// creation of a [`Checksum`].
+
+pub struct GgrsChecksumPlugin;
+
+impl GgrsChecksumPlugin {
+    /// A [`System`] responsible for updating [`Checksum`] based on [`ChecksumParts`](`ChecksumPart`).
+    pub fn update(mut checksum: ResMut<Checksum>, parts: Query<&ChecksumPart>) {
+        // TODO: Add explicit ordering to `ChecksumPart`'s to make checksum more robust to transposition
```
johanhelsing, line 41: This is done now, right?

src/schedule_systems.rs
bushrat011899: Within the `SyncTest` session type there is a parameter, `depth`, which is how far back the rollback will go during testing. So for a `SyncTest`, the confirmed frame is always `frame - depth`.

gschup: We should just have a getter for that in ggrs.

src/snapshot/checksum.rs
bushrat011899: If there are two things that contribute to the checksum (lets say a resource `Foo(usize)` and a resource `Bar(usize)`), their individual hashes will be identical for identical inner values (type information isn't passed into `Hash`). So if you have `Foo(5)` and `Bar(7)` on frame 5, then `Foo(7)` and `Bar(5)` on frame 6, frames 5 and 6 will have identical checksums, because `a ^ b == b ^ a`.

By having a guaranteed ordering, you could instead hash the whole sequence, which will produce unique checksums for frames 5 and 6 in the above example.

At first I thought about including `TypeId` or `type_name` in the hash, since we have access to that information and it's order independent, but those values aren't stable yet, so it's possible the checksum could just break for cross-platform multiplayer.

A workaround for that could be to have a `HashMap<TypeId, usize>` that's built as data types are added to the checksum process, but I don't know if the above transposition issue is even common enough to justify the added overhead.

src/schedule_systems.rs
bushrat011899: > We should just have a getter for that in ggrs. I'll write it later today :)

Thanks!

johanhelsing (commented):
Read through everything now. Just a few minor things left on my part.

src/snapshot/checksum.rs
johanhelsing: Ah, I see. Makes sense.

johanhelsing: We could create a constructor where we pass in a value and leave it up to calling code. Or it could be a const generic. Just ideas. Think I prefer the simpler API over more robust checksumming.

src/snapshot/rollback_entity_map.rs
```diff
@@ -0,0 +1,42 @@
...
+/// changed during a rollback.
+#[derive(Resource, Default)]
+pub struct RollbackEntityMap(EntityMap);
+
+impl RollbackEntityMap {
+    pub fn new(map: EntityMap) -> Self {
+        Self(map)
+    }
+
+    pub fn get_map(&self) -> EntityMap {
```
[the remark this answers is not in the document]
bushrat011899: Agreed, I'll rename that function to `generate_map`, which I think better conveys the potential cost of the operation.

src/snapshot/component_checksum_hash.rs
```diff
@@ -0,0 +1,77 @@
...
+            component.hash(&mut hasher);
+
+            // XOR chosen over addition or multiplication as it is closed on u64 and commutative
+            result ^= hasher.finish();
+        }
+
+        // Hash the XOR'ed result to break commutativity with other types
+        result.hash(&mut hasher);
+
+        let result = ChecksumPart(hasher.finish() as u128);
```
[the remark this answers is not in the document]
bushrat011899: Yeah that shouldn't be hard at all. I'll add `trace!` to the component and resource checksum generators, and the final checksum generator as well.

src/snapshot/set.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// which had to be recreated could not use the same ID, so any data referring to that ID is now invalid.
+    /// Once this set completes, all data should now be coherent with the [`World`].
+    Mapping,
+    /// Flush any deferred operations
+    MappingFlush,
+}
+
+#[derive(SystemSet, Hash, Debug, PartialEq, Eq, Clone)]
+pub enum SaveWorldSet {
+    /// Generate checksums for any tracked data.
```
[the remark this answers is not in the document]
bushrat011899: Agreed!

src/snapshot/resource_checksum_hash.rs
```diff
@@ -0,0 +1,61 @@
...
+        let mut hasher = bevy::utils::FixedState.build_hasher();
+
+        resource.hash(&mut hasher);
+
+        let result = ChecksumPart(hasher.finish() as u128);
+
+        if let Ok(mut checksum) = checksum.get_single_mut() {
+            *checksum = result;
+        } else {
+            commands.spawn((result, ChecksumFlag::<R>::default()));
```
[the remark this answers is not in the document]
bushrat011899: That's a good point! I already setup a `SaveWorldSet::PreSnapshotFlush`, so I'll move the `Checksum` update function to run after the flush but before snapshot itself.

src/lib.rs
```diff
@@ -90,8 +89,9 @@ impl Default for FixedTimestepData {
 #[derive(Resource, Debug, Default)]
 pub struct RollbackFrameCount(i32);
 
-#[derive(Resource)]
-struct RollbackTypeRegistry(TypeRegistryInternal);
+/// The most recently confirmed frame. Any information for frames stored before this point can be safely discarded.
+#[derive(Resource, Debug, Default)]
+pub struct ConfirmedFrameCount(i32);
```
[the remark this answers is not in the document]
bushrat011899: I've added a bunch of sensible traits to both, and then implemented `Into<i32>` as an access method. That should be enough for most people to work with these resources without too much hassle.

examples/box_game/box_game_spectator.rs
```diff
@@ -41,10 +41,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .set_rollback_schedule_fps(FPS)
         // this system will be executed as part of input reading
         .add_systems(ReadInputs, read_local_inputs)
-        // register types of components AND resources you want to be rolled back
-        .register_rollback_component::<Transform>()
-        .register_rollback_component::<Velocity>()
-        .register_rollback_resource::<FrameCount>()
+        // Rollback behavior can be controlled through many modular plugins
```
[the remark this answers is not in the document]
bushrat011899: Good point, I'll update the wording to focus on the extension methods, but mention the plugins as well for those that don't have methods (yet?)

src/snapshot/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+    depth: usize,
+    _phantom: PhantomData<For>,
+}
+
+impl<For, As> Default for GgrsSnapshots<For, As> {
+    fn default() -> Self {
+        Self {
+            snapshots: VecDeque::with_capacity(DEFAULT_FPS),
+            frames: VecDeque::with_capacity(DEFAULT_FPS),
+            depth: DEFAULT_FPS, // TODO: Make sensible choice here
```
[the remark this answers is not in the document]
bushrat011899: It definitely should be decided you're right. I think `DEFAULT_FPS` is a good default value, since I think it's important that snapshots not be collected infinitely if (somehow) a confirmed frame is never reached. But I'm open to other opinions on this one, as I have somewhat arbitrarily just decided that 1 second of rollback is a good maximum.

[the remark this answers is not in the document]
gschup: It is impossible for GGRS to step ahead of the confirmed frame by more than `max_prediction_window` (parameter in the session, default is 8). You should use the parameter from the session to construct this.

gschup:
I have started looking into the code, but I recently started a new job, so my review progress might be slow :)

bushrat011899:
> I have started looking into the code

Woo!

> but I recently started a new job, so my review progress might be slow :)

Congratulations! 🎉

src/snapshot/mod.rs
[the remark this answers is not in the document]
bushrat011899: Of course that makes much more sense! I've set the default value to 8, and I've updated the `discard_old_snapshots` system to also check if the max prediction window has changed, and use that as their depth.

src/schedule_systems.rs
gschup: [It now exists](https://github.com/gschup/ggrs/commit/d062eadaba25ac0b77c85d61ebca032e7c32bc05).

src/lib.rs
```diff
@@ -182,61 +149,124 @@ impl<C: Config> Plugin for GgrsPlugin<C> {
...
-            .add_systems(SaveWorld, save_world);
+            .add_plugins((
+                GgrsSnapshotSetPlugin,
+                GgrsChecksumPlugin,
+                GgrsResourceSnapshotCopyPlugin::<Checksum>::default(),
+                GgrsEntitySnapshotPlugin,
+                GgrsComponentSnapshotReflectPlugin::<Parent>::default(),
+                GgrsComponentMapEntitiesPlugin::<Parent>::default(),
+                GgrsComponentSnapshotReflectPlugin::<Children>::default(),
+                GgrsComponentMapEntitiesPlugin::<Children>::default(),
```
gschup, line 169: The names for these plugins are very verbose und hard to read. Is the user ever going to touch them? If so, they probably need a nicer name. Even if not, it still should be a consideration.

src/lib.rs
```diff
@@ -182,61 +149,124 @@ impl<C: Config> Plugin for GgrsPlugin<C> {
...
+    fn rollback_resource_with_clone<Type>(&mut self) -> &mut Self
+    where
+        Type: Resource + Clone;
+
+    /// Registers a component type for saving and loading from the world. This
+    /// uses [`reflection`](`Reflect`) based snapshots for rollback.
+    ///
+    /// NOTE: Unlike previous versions of `bevy_ggrs`, this will no longer automatically
+    /// apply entity mapping through the [`MapEntities`](`bevy::ecs::entity::MapEntities`) trait.
+    /// If you require this behavior, see [`GgrsComponentMapEntitiesPlugin`].
```
gschup, line 205: You hint towards the documentation of this plugin, but I cannot find any helpful text there that would further educate the user how to deal with this.

In addition to extra documentation, we probably need a parented example (e.g. cube with smaller child cube on top) that showcases usage of this.

src/snapshot/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            self.frames.len(),
+            "Snapshot and Frame queues must always be in sync"
+        );
+
+        loop {
+            let Some(&current) = self.frames.front() else {
+                break;
+            };
+
+            // TODO: Handle wrapping behavior (wrapping_sub, etc.)
```
gschup, line 117: is this still an open todo? what exactly is meant here?

src/world_snapshot.rs
gschup, line 1: good riddance, away with ye!

src/snapshot/mod.rs
bushrat011899: Yes I missed that! I wanted to add some simple logic to handle the possibility of the `i32` frame count wrapping. I know that's a highly unlikely event to actually occur during gameplay, but I still wanted to make sure it was handled.

src/lib.rs
bushrat011899: I agree, I've gone over all the plugins and added some extra documentation and at least 1 example demonstrating how to use the plugin. As a benefit, I've ensured every example properly compiles as a doc-test, so this has also increased test coverage.

As for the example, I completely agree. I think it would be good to maybe work on a stress-test example, as a benchmark. Since, as far as I can tell, there currently isn't really a benchmark in this crate for testing performance regression against.

src/lib.rs
bushrat011899: I agree, they're way too verbose. I've completed the set of extension methods on `App` to allow an end user to not use these plugins by name at all.

bushrat011899:
To spur discussion, I've started profiling this PR versus the current main branch, just to see what (if any) performance changes this would introduce. Running the current sync test example on my laptop, this PR **increases** running time for the `run_ggrs_schedules` system by approximately 1ms (from 2.3ms to 3.3ms). I suspect almost all of that change is the additional overhead in running many smaller systems instead of a handful of large ones. I will look into creating a proper benchmark game with far more entities, components, etc. to compare against.

My hypothesis is that for a more realistic application, the added overhead will be outweighed by the ability to run systems in parallel. Regardless of whether that theory holds or not, I will also be spending the next couple of days seeing if I can tune the performance.

bushrat011899:
Once I got home I did some more testing, this appears to have a _massive_ performance boost for large quantities of rollback components like I suspected. On my laptop I ran the following:

```bash
cargo run --example box_game_synctest --release --features bevy/trace_chrome -- --num-players 1000 --check-distance 5
```

- On main this takes an average of 1.2ms to complete the `LoadWorld` schedule, and 2.7ms to complete the `SaveWorld`.
- On this PR, it takes 3.2ms for `LoadWorld` and only 0.8ms for `SaveWorld`, _for 10,000 players_!

On the main branch, I can't even run above 1,000 players, it just crashes on startup. If I instead run this PR with "only" 1,000 players, loading and saving both drop down to 0.26ms. Failure to run doesn't occur until above 15,000 players. Tomorrow I'm going to try producing some charts to visualise the scaling difference. Based on this rudimentary testing tho, it appears my hunch was correct in that the overhead of additional save/load systems is totally negligible once a significant number of rollback entities and data is at play.

gschup:
> On the main branch, I can't even run above 1,000 players, it just crashes on startup. If I instead run this PR with "only" 1,000 players, loading and saving both drop down to 0.26ms. Failure to run doesn't occur until above 15,000 players. [...]

What exactly do you mean by player? Cubes per player? In any case, very cool and promising results :)

bushrat011899:
> What exactly do you mean by player? Cubes per player? In any case, very cool and promising results :)

Leaving the box game test unchanged and setting the command line `numplayers` option. So strictly speaking, 10,000 synctest players. I would fully expect such a test to be worthless for actual networked play, as I don't think it's be feasible to ever have 10,000 P2P sessions active (60fps, minimum of 1 byte per input packet, 10,000 packets RX and TX, totals 120KBps on its own, completely ignoring all the other overhead involved)

src/lib.rs
gschup: I think it could be sensible to just drop the Ggrs prefix. Such a shame that Components and Resources need different handling and thus different plugins, otherwise, we could drop that part of the name, too :)

bushrat011899:
# Performance Testing

## Equipment

- AMD Ryzen R9 5950X limited to 65W TDP and 8 cores
- 64GB of RAM
- Windows 10

## Procedure

- Ran `cargo run --example box_game_synctest --release --features bevy/trace_chrome -- --check-distance 5 --num-players {X}` for various `X` values.
- Increased `X` until game no longer plays (crash on launch, immediate freeze, etc.)
- Used [Perfetto](https://ui.perfetto.dev) to get average run time for `PreUpdate`, `LoadWorld`, and `SaveWorld` schedules.
- Collated results in Excel.

## Results

Main
PR
_Note that the vertical and horizontal axes are both logarithmic (expecting a doubling of entities to double runtime, therefore should be linear on log-log). Additionally, the vertical axis is execution time in seconds, while the horizontal axis is the `num-players` parameter_

## Interpretation

This PR improves performance for the box game sync-test example where there are more than 10 players, by an order of magnitude. Since 10 rollback entities is (in my opinion) an expected number of entities to include in a typical multiplayer game, it is reasonable to state that this PR improves performance in all expected cases.

johanhelsing:
I had a go at making a stress test example #78. Can be cherry-picked on top of this PR.

An interesting finding is that I seem to be dropping a lot of frames when spawning a bunch of new particles, but then it runs smoothly afterwards.

I also made branch to run on main for comparison: https://github.com/gschup/bevy_ggrs/compare/main...johanhelsing:bevy_ggrs:particles-stress-test-old?expand=1

src/snapshot/component_reflect.rs
```diff
@@ -72,6 +80,12 @@ where
                 (None, None) => {}
             }
         }
+
+        trace!(
+            "Rolled Back {} {} component(s)",
```
johanhelsing, line 105:
```suggestion
            "Rolled back {} {} component(s)",
```

repeats

johanhelsing (approved):
This is already a huge improvement both in terms of extensibility and performance.

As far as I'm concerned, this is good to go :)

We can always pursue further performance improvements in other PRs.

src/rollback.rs
```diff
@@ -45,3 +52,37 @@ impl<'w, 's, 'a> AddRollbackCommandExtension for EntityCommands<'w, 's, 'a> {
...
+pub struct RollbackOrdered {
+    order: HashMap<Rollback, usize>,
+    sorted: Vec<Rollback>,
+}
+
+impl RollbackOrdered {
+    /// Register a new [`Rollback`] for explicit ordering.
+    fn push(&mut self, rollback: Rollback) -> &mut Self {
+        self.sorted.push(rollback);
+        self.sorted.sort_unstable();
```
johanhelsing, line 67: Items are never removed from these collections, which means as the game progresses and more and more entities have been added and despawned, spawning new entities will not only leak memory, but become more expensive as the cost of the sort grows.

src/lib.rs
johanhelsing: Just a thought, but maybe it would be possible to make an API like this:

```rust
app.rollback_components_with_clone::<(Transform, Sprite, Velocity)>();
```

bushrat011899: I believe that is possible. We'd likely need a `impl_tuple` macro to avoid a extreme amount of repetition, but Bevy has several examples of exactly that macro anyway.

src/rollback.rs
bushrat011899: Good point. I've replaced the `sort_unstable` with a back-first bubble sort. Since `Rollback` entities should be created in order anyway (the first `Rollback` will be for the first entity, etc.), this should entirely avoid sorting in the vast majority of cases, and only take `n` swaps in the worst case.

As for garbage collecting dead `Rollback`'s out of this collection, I do agree that it would be a good idea, but I'm not sure what the best solution here is. I think it would be to compare the the `Rollback`'s contained in snapshots against `RollbackOrdered` and remove any that are no longer referenced, but I think that process would be fairly expensive to run every frame.

Since a pair of `Rollback` values (one in `sorted` and another in `order`) and an index as a `usize` take less than 24 bytes, it would take over 40,000 rollback entities before this object would hit 1MB of memory consumption, so the rate of lost memory would be incredibly slow.

bushrat011899:
Ok I'm done tinkering with this PR now. I made one last change where I've created a `Strategy` trait, allowing for completely generic `ComponentSnapshot` and `ResourceSnapshot` plugins:

```rust
app.add_plugins((
    ComponentSnapshotPlugin::<ReflectStrategy<Parent>>::default(),
    ComponentSnapshotPlugin::<CloneStrategy<Foo>>::default(),
    ResourceSnapshotPlugin::<CopyStrategy<Bar>>::default(),
));
```

A `Strategy` just describes how to load/store/update a snapshot:

```rust
/// A [`Strategy`] based on [`Clone`]
pub struct CloneStrategy<T: Clone>(PhantomData<T>);

impl<T: Clone> Strategy for CloneStrategy<T> {
    type Target = T;

    type Stored = T;

    #[inline(always)]
    fn store(target: &Self::Target) -> Self::Stored {
        target.clone()
    }

    #[inline(always)]
    fn load(stored: &Self::Stored) -> Self::Target {
        stored.clone()
    }
}
```

bushrat011899:
Whoops! I tidied it to death, I'll fix that import issue once I'm back home!

gschup:
Really nice idea with the Strategy!

gschup (approved):
After previous reviewers did such a great job, it's hard to find any glaring issues! This really feels like a giant leap forward from the ugly chuck of code that was `world_snapshot.rs`. Thank you so much for the work!

bushrat011899:
> After previous reviewers did such a great job, it's hard to find any glaring issues! This really feels like a giant leap forward from the ugly chunk of code that was `world_snapshot.rs`.

Thank you very much for saying that, it really means a lot!

> Thank you so much for the work!

No trouble at all, it was my pleasure! And thanks to yourself, @johanhelsing, and @nezuo for the help along the way!

PR #77, opened 2023-10-25
Switch from `structopt` to `clap` in examples

johanhelsing:
Structopt is in maintenance mode.


johanhelsing:
Oops, this changes behavior slightly. This doesn't work anymore

```
cargo run --example box_game_p2p -- --local-port 7000 --players localhost 127.0.0.1:7001
```

this does:

```
cargo run --example box_game_p2p -- --local-port 7001 --players 127.0.0.1:7000 --players localhost
```

...but I'll see if I can get the behavior to stay the same with clap

gschup:
good to know structopt is no longer a thing. Thanks!

PR #78, opened 2023-10-25
Add particles stress test

johanhelsing:
## Objective

We need some easy way of checking for performance regressions.

## Solution

- Add an example that spawns particles when either of the players holds down a key.
- Pass most parameters through command-line args

Depends on #76 and #77 

## todo:

- [x] checksumming
- [x] figure out why it desyncs
- [x] command-line options for using reflect-based snapshotting instead?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ ggrs = { git = "https://github.com/gschup/ggrs", features=["sync-send"]}
 [dev-dependencies]
 clap = { version = "4.4", features = ["derive"] }
 rand = "0.8.4"
+rand_xoshiro = "0.6"
 bevy = "0.11"
 serde = "1.0.130"
 serde_json = "1.0"
@@ -43,3 +44,7 @@ path = "examples/box_game/box_game_spectator.rs"
 [[example]]
 name = "box_game_synctest"
 path = "examples/box_game/box_game_synctest.rs"
+
+[[example]]
+name = "particles"
+path = "examples/stress_tests/particles.rs"
--- a/examples/stress_tests/particles.rs
+++ b/examples/stress_tests/particles.rs
@@ -3,7 +3,7 @@ use bevy_ggrs::{prelude::*, LocalInputs, LocalPlayers};
 use clap::Parser;
 use ggrs::UdpNonBlockingSocket;
 use rand::{Rng, SeedableRng};
-use std::net::SocketAddr;
+use std::{hash::Hasher, net::SocketAddr};
 
 #[derive(Parser, Resource)]
 struct Args {
@@ -64,6 +64,14 @@ fn read_local_inputs(
 #[derive(Default, Reflect, Component, Clone, Copy, Deref, DerefMut)]
 struct Velocity(Vec3);
 
+impl std::hash::Hash for Velocity {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.0.x.to_bits().hash(state);
+        self.0.y.to_bits().hash(state);
+        self.0.z.to_bits().hash(state);
+    }
+}
+
 #[derive(Default, Reflect, Component, Clone, Copy, Deref, DerefMut)]
 struct Ttl(usize);
 
@@ -117,6 +125,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .rollback_component_with_copy::<Velocity>()
         .rollback_component_with_copy::<Ttl>()
         .rollback_resource_with_clone::<ParticleRng>()
+        .checksum_component_with_hash::<Velocity>()
+        // todo: ideally we'd also be doing checksums for Transforms, but that's
+        // currently very clunky to do.
         .insert_resource(args)
         .add_plugins(DefaultPlugins.set(WindowPlugin {
             primary_window: Some(Window {
--- a/examples/stress_tests/particles.rs
+++ b/examples/stress_tests/particles.rs
@@ -1,0 +1,212 @@
+use bevy::{math::vec3, prelude::*, utils::HashMap, window::WindowResolution};
+use bevy_ggrs::{prelude::*, LocalInputs, LocalPlayers};
+use clap::Parser;
+use ggrs::UdpNonBlockingSocket;
+use rand::{Rng, SeedableRng};
+use std::net::SocketAddr;
+
+#[derive(Parser, Resource)]
+struct Args {
+    #[clap(short, long)]
+    local_port: u16,
+    #[clap(short, long, num_args = 1..)]
+    players: Vec<String>,
+    #[clap(short, long, num_args = 1..)]
+    spectators: Vec<SocketAddr>,
+    #[clap(short, long, default_value = "2")]
+    input_delay: usize,
+    #[clap(short, long, default_value = "10")]
+    desync_detection_interval: u32,
+    #[clap(short = 'n', long, default_value = "100")]
+    rate: u32,
+    #[clap(short, long, default_value = "60")]
+    fps: usize,
+    #[clap(long, default_value = "8")]
+    max_prediction: usize,
+}
+
+type Config = GgrsConfig<u8>;
+
+const INPUT_UP: u8 = 1 << 0;
+const INPUT_DOWN: u8 = 1 << 1;
+const INPUT_LEFT: u8 = 1 << 2;
+const INPUT_RIGHT: u8 = 1 << 3;
+
+fn read_local_inputs(
+    mut commands: Commands,
+    keyboard_input: Res<Input<KeyCode>>,
+    local_players: Res<LocalPlayers>,
+) {
+    let mut local_inputs = HashMap::new();
+
+    for handle in &local_players.0 {
+        let mut input: u8 = 0;
+
+        if keyboard_input.pressed(KeyCode::W) {
+            input |= INPUT_UP;
+        }
+        if keyboard_input.pressed(KeyCode::A) {
+            input |= INPUT_LEFT;
+        }
+        if keyboard_input.pressed(KeyCode::S) {
+            input |= INPUT_DOWN;
+        }
+        if keyboard_input.pressed(KeyCode::D) {
+            input |= INPUT_RIGHT;
+        }
+
+        local_inputs.insert(*handle, input);
+    }
+
+    commands.insert_resource(LocalInputs::<Config>(local_inputs));
+}
+
+#[derive(Default, Reflect, Component, Clone, Copy, Deref, DerefMut)]
+struct Velocity(Vec3);
+
+#[derive(Default, Reflect, Component, Clone, Copy, Deref, DerefMut)]
+struct Ttl(usize);
+
+type GameRng = rand_xoshiro::Xoshiro256PlusPlus;
+
+#[derive(Resource, Component, Clone, Deref, DerefMut)]
+struct ParticleRng(GameRng);
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let args = Args::parse();
+    let num_players = args.players.len();
+    assert!(num_players > 0);
+
+    let mut session_builder = SessionBuilder::<Config>::new()
+        .with_num_players(num_players)
+        .with_desync_detection_mode(ggrs::DesyncDetection::On {
+            interval: args.desync_detection_interval,
+        })
+        .with_max_prediction_window(args.max_prediction)?
+        .with_input_delay(args.input_delay);
+
+    for (i, player_addr) in args.players.iter().enumerate() {
+        if player_addr == "localhost" {
+            session_builder = session_builder.add_player(PlayerType::Local, i)?;
+        } else {
+            let remote_addr: SocketAddr = player_addr.parse()?;
+            session_builder = session_builder.add_player(PlayerType::Remote(remote_addr), i)?;
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


bushrat011899:
This looks great! Once #76 is merged I'll have a more detailed look into how this benchmark works.

bushrat011899:
I ran some profiling and I can see where the slowdown is coming from, the `AddRollbackCommand` needs exclusive world access, and each call needed to sort the whole list of `Rollback` components [as you pointed out](https://github.com/gschup/bevy_ggrs/pull/76#discussion_r1372424132). I upstreamed this PR to include the changes I have since made and now that lag-spike appears to be non-existent.

johanhelsing:
I added checksumming to `Velocity`, and unfortunately this example desyncs if I spawn quite a lot of particles.

I've also been experimenting a bit with [bevy_gaff](https://github.com/johanhelsing/bevy_gaff) on this version. I fixed a bunch of non-determinism issues in bevy_xpbd (locally), but I'm still left with some strange rare desyncs there that I don't understand. So I'm thinking perhaps we still have a bug in bevy_ggrs?

johanhelsing:
Another interesting thing is that it only seems to happen if rollbacks actually spawn new particles. I added an `N` key for a no-op, it doesn't actually change anything, but still triggers a rollback because the input is different, and remapped spawning to the space key.

- Releasing space doesn't trigger desyncs
- Neither does pressing or releasing N
- Pushing down space usually trigger desyncs, but not always

Cargo.toml
```diff
@@ -43,3 +44,7 @@ path = "examples/box_game/box_game_spectator.rs"
 [[example]]
 name = "box_game_synctest"
 path = "examples/box_game/box_game_synctest.rs"
+
+[[example]]
```
bushrat011899, line 48: It would be nice to add a short blurb in the `README.md` explaining how to run this stress test:

```bash
cargo run --release --example particles -- --local-port 7000 --players localhost 127.0.0.1:7001 --input-delay 0 --desync-detection-interval 1 --rate 1
cargo run --release --example particles -- --local-port 7001 --players 127.0.0.1:7000 localhost --input-delay 0 --desync-detection-interval 1 --rate 1
```

examples/stress_tests/particles.rs
```diff
@@ -0,0 +1,226 @@
+use bevy::{math::vec3, prelude::*, utils::HashMap, window::WindowResolution};
+use bevy_ggrs::{prelude::*, LocalInputs, LocalPlayers};
+use clap::Parser;
```
bushrat011899, line 3: Good choice, I think `clap` is fantastic.

examples/stress_tests/particles.rs
```diff
@@ -0,0 +1,226 @@
...
+        .rollback_component_with_clone::<GlobalTransform>()
+        .rollback_component_with_clone::<Handle<Image>>()
+        .rollback_component_with_clone::<Visibility>()
+        .rollback_component_with_clone::<ComputedVisibility>()
+        // Also add our own types
+        .rollback_component_with_copy::<Velocity>()
+        .rollback_component_with_copy::<Ttl>()
+        .rollback_resource_with_clone::<ParticleRng>()
+        .checksum_component_with_hash::<Velocity>()
+        // todo: ideally we'd also be doing checksums for Transforms, but that's
```
bushrat011899, line 127: Agreed. Perhaps a `ComponentChecksumCustomHashPlugin` which would accept a function `fn<H: Hasher>(&C, &mut H)` could be used to retroactively add hashing to types which don't support it?

examples/stress_tests/particles.rs
```diff
@@ -0,0 +1,226 @@
...
+                },
+                Velocity(vec3(rng.gen_range(-s..s), rng.gen_range(-s..s), 0.0)),
+                Ttl(ttl),
+            ))
+            .add_rollback();
+    }
+}
+
+fn update_particles(mut particles: Query<(&mut Transform, &mut Velocity)>, args: Res<Args>) {
+    let time_step = 1.0 / args.fps as f32; // todo: replace with bevy_ggrs resource?
```
bushrat011899, line 188: On this theme, I think it would be good to replace the [Time](https://docs.rs/bevy/latest/bevy/time/struct.Time.html) resource during advance-frame with one that `bevy_ggrs` manually controls the time of. Then, systems already using the canonical Bevy `Time` resource would just automatically be rollback aware.

examples/stress_tests/particles.rs
```diff
@@ -0,0 +1,226 @@
...
+                match event {
+                    GgrsEvent::Disconnected { .. } | GgrsEvent::NetworkInterrupted { .. } => {
+                        warn!("GGRS event: {event:?}")
+                    }
+                    GgrsEvent::DesyncDetected {
+                        local_checksum,
+                        remote_checksum,
+                        frame,
+                        ..
+                    } => panic!("Desync on frame {frame}. Local checksum: {local_checksum:X}, remote checksum: {remote_checksum:X}"),
```
bushrat011899, line 219: I think having a command line option to choose between `error!` and `panic!` on desync would be nice. Switching this to `error!`, I noticed that while the checksums differ, the actual simulations look identical, so whatever desync there is, it's not noticeable without the checksum.

bushrat011899 (commented):
Really like this stress test so far! It's highlighted some very interesting behaviour regarding desync detection and helped me solve at least 1 performance regression already (`sort_unchecked` in the previous PR)

examples/stress_tests/particles.rs
johanhelsing: Good idea. I noticed I changed this back and forth a number of times when debugging.

And good spot about the visual similarity.

bushrat011899:
>     * Releasing space doesn't trigger desyncs

I'm still testing with this trying to work out what's going wrong and, unfortunately, I've seen it trigger desyncs purely on the release of spacebar. I've also tested using a much coarser hashing function (only hash the integer component of velocity) with no difference in desync behaviour. I suspect the issue is related to the creation or removal of rollback entities on predicted frames. I want to resolve this, since I see it as a real blocker to more advanced future functionality (e.g., de-sync resolution through resynchronisation, etc.)

bushrat011899:
Ok it is absolutely to do with entities only and has *nothing* to do with the velocity component. I created an `EntityChecksumPlugin`:

```rust
use std::hash::{BuildHasher, Hash, Hasher};

use bevy::prelude::*;

use crate::{ChecksumFlag, ChecksumPart, Rollback, RollbackOrdered, SaveWorld, SaveWorldSet};

pub struct EntityChecksumPlugin;

impl EntityChecksumPlugin {
    #[allow(clippy::type_complexity)]
    pub fn update(
        mut commands: Commands,
        rollback_ordered: Res<RollbackOrdered>,
        components: Query<&Rollback, (With<Rollback>, Without<ChecksumFlag<Entity>>)>,
        mut checksum: Query<&mut ChecksumPart, (Without<Rollback>, With<ChecksumFlag<Entity>>)>,
    ) {
        let mut hasher = bevy::utils::FixedState.build_hasher();

        let mut result = 0;

        for &rollback in components.iter() {
            let mut hasher = hasher.clone();

            // Hashing the rollback index ensures this hash is unique and stable
            rollback_ordered.order(rollback).hash(&mut hasher);

            // XOR chosen over addition or multiplication as it is closed on u64 and commutative
            result ^= hasher.finish();
        }

        // Hash the XOR'ed result to break commutativity with other types
        result.hash(&mut hasher);

        let result = ChecksumPart(hasher.finish() as u128);

        trace!(
            "Rollback Entities have checksum {:X}",
            result.0
        );

        if let Ok(mut checksum) = checksum.get_single_mut() {
            *checksum = result;
        } else {
            commands.spawn((result, ChecksumFlag::<Entity>::default()));
        }
    }
}

impl Plugin for EntityChecksumPlugin {
    fn build(&self, app: &mut App) {
        app.add_systems(SaveWorld, Self::update.in_set(SaveWorldSet::Checksum));
    }
}
```

And added it as the only system contributing to the checksum and the same desync behaviour is observed. I'm about to go to bed but that leaves 3 possibilities:

1. Particles are being spawned when they shouldn't be.
2. Particles aren't being despawned when they should be.
3. The value in `RollbackOrdered` is not synchronised.

Writing this out, I now know it has to be 3. I think if a rollback entity is added on a predicted frame, and then is rolled back, it'll stay in the order resource and break sync with all clients. In the morning I'll make a PR that just ensures rollback entities created pre-emptively are removed from `RollbackOrdered`.

bushrat011899:
I believe I have the desync issues with this branch solved with #82! It was a combination of what I suspected (rollback entities being de/spawned during a rollback messing up the `RollbackOrdered` ordering) and a completely different issue where Bevy would chose a different `generation` value for entities spawned during a rollback, destroying `Rollback` ordering for all clients that had to rollback.

I tested with a rebase and couldn't induce a desync. I'm honestly shocked that hashing a `f32` velocity value by its bits _wasn't_ the cause of the desync issues!

johanhelsing:
> I'm honestly shocked that hashing a `f32` velocity value by its bits _wasn't_ the cause of the desync issues!

As far as I've understood, it's fine as long as we don't get any NaN values, in fact we rely on it for determinisim. I guess the hasher could check for NaN and just do a deterministic bogus checksum (or warn)?

gschup:
We should definitely avoid hashing the bits of a float just like that. Special cases like NaN (both quiet and signalling) and Inf can become troublesome.

https://reference.opcfoundation.org/Core/Part6/v104/docs/5.2.2.3#:~:text=The%20floating%2Dpoint%20type%20supports,000000000000F8FF)%20or%20(0000C0FF).

gschup:
On the other hand, working with floats in Rust has proven to be surprisingly deterministic in my experience.

johanhelsing:
Btw, this sho

> We should definitely avoid hashing the bits of a float just like that. Special cases like NaN (both quiet and signalling) and Inf can become troublesome.
> 
> https://reference.opcfoundation.org/Core/Part6/v104/docs/5.2.2.3#:~:text=The%20floating%2Dpoint%20type%20supports,000000000000F8FF)%20or%20(0000C0FF).

I'm of two minds... aren't you in a bit of a pickle if your floats have reached inf or nan anyway? Wouldn't it be good to know?

gschup:
> I'm of two minds... aren't you in a bit of a pickle if your floats have reached inf or nan anyway? Wouldn't it be good to know?

For our purposes, definitely. I think adding a NaN/Inf check just before hashing the bits should be sufficient for this example.

gschup:
> I'm of two minds... aren't you in a bit of a pickle if your floats have reached inf or nan anyway? Wouldn't it be good to know?

For our purposes, definitely. I think adding a NaN/Inf check just before hashing the bits should be sufficient for this example.

Cargo.toml
johanhelsing: I put in a clap help section instead, which I think makes it easy to discover (and keep in sync, since it's in the source file).

```
dev/bevy_ggrs particles-stress-test↑1 ~1 →1 …1 cargo run --release --example particles
    Finished release [optimized] target(s) in 0.31s
     Running `target\release\examples\particles.exe`
error: the following required arguments were not provided:
  --local-port <LOCAL_PORT>

Usage: particles.exe --local-port <LOCAL_PORT>

For more information, try '--help'.
error: process didn't exit successfully: `target\release\examples\particles.exe` (exit code: 2)
dev/bevy_ggrs particles-stress-test↑1 ~1 →1 …1 cargo run --release --example particles -- --help
    Finished release [optimized] target(s) in 0.30s
     Running `target\release\examples\particles.exe --help`
Stress test for bevy_ggrs

## Basic usage:

Player 1:

cargo run --release --example particles -- --local-port 7000 --players localhost 127.0.0.1:7001

Player 2:

cargo run --release --example particles -- --local-port 7001 --players 127.0.0.1:7001 localhost

Usage: particles.exe [OPTIONS] --local-port <LOCAL_PORT>

Options:
  -l, --local-port <LOCAL_PORT>
          The udp port to bind to for this peer

  -p, --players <PLAYERS>...
          Address and port for the players. Order is significant. Put yourself as "localhost".

          e.g. `--players localhost 127.0.0.1:7001`

  -s, --spectators <SPECTATORS>...
          Address and port for any spectators

  -i, --input-delay <INPUT_DELAY>
          How long inputs should be kept before they are deployed. A low value, such as 0 will result in low latency, but plenty of rollbacks

          [default: 2]

  -d, --desync-detection-interval <DESYNC_DETECTION_INTERVAL>
          How often the clients should exchange and compare checkums of state

          [default: 10]

      --continue-after-desync
          Whether to continue after a detected desync, the default is to panic

  -n, --rate <RATE>
          How many particles to spawn per frame

          [default: 100]

  -f, --fps <FPS>
          Simulation frame rate

          [default: 60]

      --max-prediction <MAX_PREDICTION>
          How far ahead we should simulate when we don't get any input from a player

          [default: 8]

      --reflect
          Whether to use reflect-based rollback. This is much slower than the default clone/copy-based rollback

  -h, --help
          Print help (see a summary with '-h')
dev/bevy_ggrs particles-stress-test↑1 ~1 →1 …1
```

johanhelsing:
> For our purposes, definitely. I think adding a NaN/Inf check just before hashing the bits should be sufficient for this example.

I added an `assert` and a comment.

gschup:
So this stress test is ready to be merged, then?

PR #80, opened 2023-10-29
Added `GGRSTimePlugin` to Provide Rollback-Safe `Time`

bushrat011899:
# Objective

Systems running in the `GgrsSchedule` currently need to be made aware of the framerate for any time dependent behaviour. This isn't ideal, as either:

- Systems need to have constants written in a per-frame unit (meters / frame for a velocity, etc.); or
- Users need to calculate a delta-time constant to allow more typical delta-time algorithms.

Bevy already provides a [`Time`](https://docs.rs/bevy/latest/bevy/time/struct.Time.html) resource which has convenience methods for getting a delta-time value, but is in real-time, and cannot be made aware of the rollback system without breaking other unrelated systems (e.g., animation players, audio systems, etc.). It would be great to have access to a rollback safe version of this resource.

Additionally, since the release of Bevy 0.12, it is now possible to have custom `Time` resources that can replace a contextual default, `Time<()>`. It would be ideal for Bevy GGRS to provide such a custom time resource which is rollback aware.

# Solution

I have created a plugin, `GGRSTimePlugin` which is included with `GgrsPlugin`. This plugin:

- Manages a `Time<GGRSTime>` resource.
- Replaces the Bevy `Time<()>` resource at the beginning of the `GgrsSchedule`, and restores the `Time<Virtual>` resource at the end.
- Snapshots `Time<GGRSTime>` to allow for rollback.

To ensure all clients have synchronised `Time<GGRSTime>` resources, it is updated on each frame to reflect that exactly a frame-period (`1/fps`) has passed. However, to avoid accumulating error, the time passed is calculated each frame as `frame/fps`, in nanoseconds:

```rust
pub fn update(
    mut time: ResMut<Time<GGRSTime>>,
    framerate: Res<RollbackFrameRate>,
    frame: Res<RollbackFrameCount>,
) {
    let this_frame = frame.0 as u64;
    let framerate = framerate.0 as u64;

    // 1_000_000_000 fits within a u32, and so does frame, making their product at most u64 in size
    // By scaling to nanoseconds, rounding error should be minimised.
    let runtime = Duration::from_nanos(this_frame * 1_000_000_000 / framerate);

    time.advance_to(runtime);
}
```

Because this method recalculates the total runtime each frame, the exact delta-time value varies frame-to-frame. But, because it is computed using the exact formula on every client, this value will be stable for all clients:

```bash
2023-10-29T11:31:02.225467Z  INFO box_game_synctest::box_game: Delta Time: 0.016666668
2023-10-29T11:31:02.225720Z  INFO box_game_synctest::box_game: Delta Time: 0.016666666
2023-10-29T11:31:02.225967Z  INFO box_game_synctest::box_game: Delta Time: 0.016666668
2023-10-29T11:31:02.226199Z  INFO box_game_synctest::box_game: Delta Time: 0.016666668
2023-10-29T11:31:02.241703Z  INFO box_game_synctest::box_game: Delta Time: 0.016666668
2023-10-29T11:31:02.242148Z  INFO box_game_synctest::box_game: Delta Time: 0.016666668
2023-10-29T11:31:02.242432Z  INFO box_game_synctest::box_game: Delta Time: 0.016666666
2023-10-29T11:31:02.242727Z  INFO box_game_synctest::box_game: Delta Time: 0.016666668
2023-10-29T11:31:02.242989Z  INFO box_game_synctest::box_game: Delta Time: 0.016666668
2023-10-29T11:31:02.243245Z  INFO box_game_synctest::box_game: Delta Time: 0.016666666
```

# Usage

Users can now access the Bevy `Time` resource in their systems safely:

```rust
pub fn get_time(time: Res<Time>) {
    // Total time elapsed since the session began
    let t = time.elapsed_seconds_f64();

    // Time passed since last frame
    let dt = time.delta().as_secs_f32();
}
```

Additionally, the `Time<GGRSTime>` resource is available in all schedules (even outside of `AdvanceWorld`/`GgrsSchedule`), which would, for example, allow UI elements to be updated to reflect the current in-game time.

```rust
pub fn get_in_game_time(game_time: Res<Time<GGRSTime>>) {
    // ...
}
```

## Tangent About dt

For those unfamiliar with delta-time based physics, having access to the amount of time that has elapsed since the last frame allows for calculating time-dependent values. For example, lets say we have a day-night cycle in our game. We'll define that an in-game day lasts 20 minutes, 10 minutes of daytime and 10 minutes of night. We can then relate in-game time to real-world time using:

```rust
// 72 in-game days per real-world day
let game_time = 72. * t;
```

We can then check if it's daytime by getting the time-of-day:

```rust
let hours = game_time / (60. * 60.);
let hour = hours % 24.;

// Sun rises at 6AM and sets at 6PM in our game
let is_daytime = hour >= 6. && hours <= 18.;
let is_nighttime = !is_daytime;
```

Having access to a delta-time value allows for many standard physics systems to be solved using finite integration:

```rust
// If we hold the up button, we'll accelerate up at 10m/s^2, otherwise we fall due to gravity at 9.81m/s^2.
let acceleration = if up_pressed { 10. } else { -9.81 };

// Using a dt value, we can approximate the change is velocity due to acceleration
velocity += acceleration  * dt;

// Likewise, we can also approximate the change is height due to velocity
height += velocity * dt;
```

The benefit of using time based systems is now our game logic is independent of the framerate chosen, leaving more options to both the developer and the end-user.

## Example

Looking at the Box Game example, the physics can then be expressed in a more natural set of operations:

```rust
pub fn move_cube_system(
    mut query: Query<(&mut Transform, &mut Velocity, &Player), With<Rollback>>,
    inputs: Res<PlayerInputs<BoxConfig>>,
    // This Time resource is rollback safe
    time: Res<Time>,
) {
    // This value is *not* constant frame-to-frame, but it is synchronized across clients.
    // This will avoid a drift in expected values when using a naive 1/fps as a constant
    let dt = time.delta().as_secs_f32();

    for (mut t, mut v, p) in query.iter_mut() {
        let input = inputs[p.handle].0.inp;
        // set velocity through key presses
        if input & INPUT_UP != 0 && input & INPUT_DOWN == 0 {
            v.z -= ACCELERATION * dt;
        }
        if input & INPUT_UP == 0 && input & INPUT_DOWN != 0 {
            v.z += ACCELERATION * dt;
        }
        if input & INPUT_LEFT != 0 && input & INPUT_RIGHT == 0 {
            v.x -= ACCELERATION * dt;
        }
        if input & INPUT_LEFT == 0 && input & INPUT_RIGHT != 0 {
            v.x += ACCELERATION * dt;
        }

        // slow down
        if input & INPUT_UP == 0 && input & INPUT_DOWN == 0 {
            v.z *= FRICTION.powf(dt);
        }
        if input & INPUT_LEFT == 0 && input & INPUT_RIGHT == 0 {
            v.x *= FRICTION.powf(dt);
        }
        v.y *= FRICTION.powf(dt);

        // constrain velocity
        let mag = (v.x * v.x + v.y * v.y + v.z * v.z).sqrt();
        if mag > MAX_SPEED {
            let factor = MAX_SPEED / mag;
            v.x *= factor;
            v.y *= factor;
            v.z *= factor;
        }

        // apply velocity
        t.translation.x += v.x * dt;
        t.translation.y += v.y * dt;
        t.translation.z += v.z * dt;

        // constrain cube to plane
        t.translation.x = t.translation.x.max(-1. * (PLANE_SIZE - CUBE_SIZE) * 0.5);
        t.translation.x = t.translation.x.min((PLANE_SIZE - CUBE_SIZE) * 0.5);
        t.translation.z = t.translation.z.max(-1. * (PLANE_SIZE - CUBE_SIZE) * 0.5);
        t.translation.z = t.translation.z.min((PLANE_SIZE - CUBE_SIZE) * 0.5);
    }
}
```

In the Box Game example, this allows for experimenting with the exact frame-rate that the game system runs at, without affecting the physics or game-play.

### DIFF
--- a/examples/box_game/box_game.rs
+++ b/examples/box_game/box_game.rs
@@ -17,9 +17,9 @@ const INPUT_DOWN: u8 = 1 << 1;
 const INPUT_LEFT: u8 = 1 << 2;
 const INPUT_RIGHT: u8 = 1 << 3;
 
-const MOVEMENT_SPEED: f32 = 0.005;
-const MAX_SPEED: f32 = 0.05;
-const FRICTION: f32 = 0.9;
+const ACCELERATION: f32 = 18.0;
+const MAX_SPEED: f32 = 3.0;
+const FRICTION: f32 = 0.0018;
 const PLANE_SIZE: f32 = 5.0;
 const CUBE_SIZE: f32 = 0.2;
 
@@ -161,31 +161,35 @@ pub fn increase_frame_system(mut frame_count: ResMut<FrameCount>) {
 pub fn move_cube_system(
     mut query: Query<(&mut Transform, &mut Velocity, &Player), With<Rollback>>,
     inputs: Res<PlayerInputs<BoxConfig>>,
+    // Thanks to RollbackTimePlugin, this is rollback safe
+    time: Res<Time>,
 ) {
+    let dt = time.delta().as_secs_f32();
+
     for (mut t, mut v, p) in query.iter_mut() {
         let input = inputs[p.handle].0.inp;
         // set velocity through key presses
         if input & INPUT_UP != 0 && input & INPUT_DOWN == 0 {
-            v.z -= MOVEMENT_SPEED;
+            v.z -= ACCELERATION * dt;
         }
         if input & INPUT_UP == 0 && input & INPUT_DOWN != 0 {
-            v.z += MOVEMENT_SPEED;
+            v.z += ACCELERATION * dt;
         }
         if input & INPUT_LEFT != 0 && input & INPUT_RIGHT == 0 {
-            v.x -= MOVEMENT_SPEED;
+            v.x -= ACCELERATION * dt;
         }
         if input & INPUT_LEFT == 0 && input & INPUT_RIGHT != 0 {
-            v.x += MOVEMENT_SPEED;
+            v.x += ACCELERATION * dt;
         }
 
         // slow down
         if input & INPUT_UP == 0 && input & INPUT_DOWN == 0 {
-            v.z *= FRICTION;
+            v.z *= FRICTION.powf(dt);
         }
         if input & INPUT_LEFT == 0 && input & INPUT_RIGHT == 0 {
-            v.x *= FRICTION;
+            v.x *= FRICTION.powf(dt);
         }
-        v.y *= FRICTION;
+        v.y *= FRICTION.powf(dt);
 
         // constrain velocity
         let mag = (v.x * v.x + v.y * v.y + v.z * v.z).sqrt();
@@ -197,9 +201,9 @@ pub fn move_cube_system(
         }
 
         // apply velocity
-        t.translation.x += v.x;
-        t.translation.y += v.y;
-        t.translation.z += v.z;
+        t.translation.x += v.x * dt;
+        t.translation.y += v.y * dt;
+        t.translation.z += v.z * dt;
 
         // constrain cube to plane
         t.translation.x = t.translation.x.max(-1. * (PLANE_SIZE - CUBE_SIZE) * 0.5);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,7 @@
 use bevy::{
     ecs::{
         entity::MapEntities,
-        schedule::{LogLevel, ScheduleBuildSettings, ScheduleLabel},
+        schedule::{ExecutorKind, LogLevel, ScheduleBuildSettings, ScheduleLabel},
     },
     prelude::*,
     utils::{Duration, HashMap},
@@ -16,10 +16,12 @@ pub use ggrs;
 
 pub use rollback::*;
 pub use snapshot::*;
+pub use time::*;
 
 pub(crate) mod rollback;
 pub(crate) mod schedule_systems;
 pub(crate) mod snapshot;
+pub(crate) mod time;
 
 pub mod prelude {
     pub use crate::{
@@ -69,9 +71,7 @@ pub enum Session<T: Config> {
 pub struct PlayerInputs<T: Config>(Vec<(T::Input, InputStatus)>);
 
 #[derive(Resource, Copy, Clone, Debug)]
-pub struct FixedTimestepData {
-    /// fixed FPS our logic is running with
-    pub fps: usize,
+struct FixedTimestepData {
     /// accumulated time. once enough time has been accumulated, an update is executed
     accumulator: Duration,
     /// boolean to see if we should run slow to let remote clients catch up
@@ -81,7 +81,6 @@ pub struct FixedTimestepData {
 impl Default for FixedTimestepData {
     fn default() -> Self {
         Self {
-            fps: DEFAULT_FPS,
             accumulator: Duration::ZERO,
             run_slow: false,
         }
@@ -132,7 +131,29 @@ pub struct LoadWorld;
 #[derive(ScheduleLabel, Debug, Hash, PartialEq, Eq, Clone)]
 pub struct SaveWorld;
 
+/// Label for the schedule which advances the current world to the next frame.
+#[derive(ScheduleLabel, Debug, Hash, PartialEq, Eq, Clone)]
+pub struct AdvanceWorld;
+
 /// GGRS plugin for bevy.
+///
+/// # Rollback
+///
+/// This will provide rollback management for the following items in the Bevy ECS:
+/// - [Entities](`Entity`)
+/// - [Parent] and [Children] components
+/// - [Time]
+///
+/// To add more data to the rollback management, see the methods provided by [GgrsApp].
+///
+/// # Rollback
+///
+/// This will provide rollback management for the following items in the Bevy ECS:
+/// - [Entities](`Entity`)
+/// - [Parent] and [Children] components
+/// - [Time]
+///
+/// To add more data to the rollback management, see the methods provided by [GgrsApp].
 ///
 /// # Examples
 /// ```rust
@@ -176,26 +197,33 @@ impl<C: Config> Default for GgrsPlugin<C> {
 
 impl<C: Config> Plugin for GgrsPlugin<C> {
     fn build(&self, app: &mut App) {
-        let mut schedule = Schedule::default();
-        schedule.set_build_settings(ScheduleBuildSettings {
-            ambiguity_detection: LogLevel::Error,
-            ..default()
-        });
-
         app.init_resource::<RollbackFrameCount>()
             .init_resource::<ConfirmedFrameCount>()
             .init_resource::<MaxPredictionWindow>()
             .init_resource::<RollbackOrdered>()
             .init_resource::<LocalPlayers>()
             .init_resource::<FixedTimestepData>()
-            .add_schedule(GgrsSchedule, schedule)
-            .add_schedule(ReadInputs, Schedule::new())
+            .init_schedule(ReadInputs)
+            .init_schedule(LoadWorld)
+            .init_schedule(SaveWorld)
+            .edit_schedule(AdvanceWorld, |schedule| {
+                // AdvanceWorld is mostly a facilitator for GgrsSchedule, so SingleThreaded avoids overhead
+                // This can be overridden if desired.
+                schedule.set_executor_kind(ExecutorKind::SingleThreaded);
+            })
+            .edit_schedule(GgrsSchedule, |schedule| {
+                schedule.set_build_settings(ScheduleBuildSettings {
+                    ambiguity_detection: LogLevel::Error,
+                    ..default()
+                });
+            })
             .add_systems(PreUpdate, schedule_systems::run_ggrs_schedules::<C>)
             .add_plugins((
                 SnapshotSetPlugin,
                 ChecksumPlugin,
                 EntitySnapshotPlugin,
                 EntityChecksumPlugin,
+                GGRSTimePlugin,
                 ResourceSnapshotPlugin::<CloneStrategy<RollbackOrdered>>::default(),
                 ComponentSnapshotPlugin::<ReflectStrategy<Parent>>::default(),
                 ComponentMapEntitiesPlugin::<Parent>::default(),
@@ -277,8 +305,7 @@ pub trait GgrsApp {
 
 impl GgrsApp for App {
     fn set_rollback_schedule_fps(&mut self, fps: usize) -> &mut Self {
-        self.world
-            .insert_resource(FixedTimestepData { fps, ..default() });
+        self.world.insert_resource(RollbackFrameRate(fps));
 
         self
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -223,7 +223,7 @@ impl<C: Config> Plugin for GgrsPlugin<C> {
                 ChecksumPlugin,
                 EntitySnapshotPlugin,
                 EntityChecksumPlugin,
-                GGRSTimePlugin,
+                GgrsTimePlugin,
                 ResourceSnapshotPlugin::<CloneStrategy<RollbackOrdered>>::default(),
                 ComponentSnapshotPlugin::<ReflectStrategy<Parent>>::default(),
                 ComponentMapEntitiesPlugin::<Parent>::default(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,7 +69,7 @@ pub enum Session<T: Config> {
 pub struct PlayerInputs<T: Config>(Vec<(T::Input, InputStatus)>);
 
 #[derive(Resource, Copy, Clone, Debug)]
-struct FixedTimestepData {
+pub struct FixedTimestepData {
     /// fixed FPS our logic is running with
     pub fps: usize,
     /// accumulated time. once enough time has been accumulated, an update is executed
--- a/src/snapshot/set.rs
+++ b/src/snapshot/set.rs
@@ -1,6 +1,6 @@
 use bevy::prelude::*;
 
-use crate::{LoadWorld, SaveWorld};
+use crate::{AdvanceWorld, GgrsSchedule, LoadWorld, SaveWorld};
 
 /// Set for ordering systems during the [`LoadWorld`] schedule.
 /// The most common option is [`LoadWorldSet::Data`], which is where [`Component`]
@@ -46,6 +46,13 @@ pub enum SaveWorldSet {
     Snapshot,
 }
 
+#[derive(SystemSet, Hash, Debug, PartialEq, Eq, Clone)]
+pub enum AdvanceWorldSet {
+    First,
+    Main,
+    Last,
+}
+
 /// Sets up the [`LoadWorldSet`] and [`SaveWorldSet`] sets, allowing for explicit ordering of
 /// rollback systems across plugins.
 pub struct SnapshotSetPlugin;
@@ -67,7 +74,32 @@ impl Plugin for SnapshotSetPlugin {
             SaveWorld,
             (SaveWorldSet::Checksum, SaveWorldSet::Snapshot).chain(),
         )
+        .configure_sets(
+            AdvanceWorld,
+            (
+                AdvanceWorldSet::First,
+                AdvanceWorldSet::Main,
+                AdvanceWorldSet::Last,
+            )
+                .chain(),
+        )
         .add_systems(LoadWorld, apply_deferred.in_set(LoadWorldSet::EntityFlush))
-        .add_systems(LoadWorld, apply_deferred.in_set(LoadWorldSet::DataFlush));
+        .add_systems(LoadWorld, apply_deferred.in_set(LoadWorldSet::DataFlush))
+        .add_systems(
+            AdvanceWorld,
+            apply_deferred
+                .after(AdvanceWorldSet::First)
+                .before(AdvanceWorldSet::Main),
+        )
+        .add_systems(
+            AdvanceWorld,
+            apply_deferred
+                .after(AdvanceWorldSet::Main)
+                .before(AdvanceWorldSet::Last),
+        )
+        .add_systems(
+            AdvanceWorld,
+            (|world: &mut World| world.run_schedule(GgrsSchedule)).in_set(AdvanceWorldSet::Main),
+        );
     }
 }
--- a/src/time.rs
+++ b/src/time.rs
@@ -18,15 +18,15 @@ impl Default for RollbackFrameRate {
 }
 
 #[derive(Default, Clone, Copy, Debug)]
-pub struct GGRSTime;
+pub struct GgrsTime;
 
-/// This plugins provides [`Time<GGRSTime>`], which is rolled-back automatically, and will also
+/// This plugins provides [`Time<GgrsTime>`], which is rolled-back automatically, and will also
 /// automatically replace [`Time<()>`] when accessed inside [`GgrsSchedule`](`crate::GgrsSchedule`).
-pub struct GGRSTimePlugin;
+pub struct GgrsTimePlugin;
 
-impl GGRSTimePlugin {
+impl GgrsTimePlugin {
     pub fn update(
-        mut time: ResMut<Time<GGRSTime>>,
+        mut time: ResMut<Time<GgrsTime>>,
         framerate: Res<RollbackFrameRate>,
         frame: Res<RollbackFrameCount>,
     ) {
@@ -42,7 +42,7 @@ impl GGRSTimePlugin {
 
     pub fn replace_default_with_ggrs(
         mut default_time: ResMut<Time<()>>,
-        ggrs_time: Res<Time<GGRSTime>>,
+        ggrs_time: Res<Time<GgrsTime>>,
     ) {
         *default_time = ggrs_time.as_generic();
     }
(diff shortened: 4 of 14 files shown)


src/time.rs
```diff
@@ -0,0 +1,135 @@
...
+        bevy_time: Option<ResMut<Time>>,
+    ) {
+        if let Some(mut bevy_time) = bevy_time {
+            let mut time = ggrs_time.time.clone();
+
+            std::mem::swap(bevy_time.as_mut(), &mut time);
+
+            ggrs_time.bevy_time = Some(time);
+        } else {
+            commands.insert_resource(ggrs_time.time.clone());
```
johanhelsing, line 101: Maybe modify the resource directly to avoid 1-frame-delay problems, and use `time.clone_from(ggrs_time.time)`?

examples/box_game/box_game_spectator.rs
```diff
@@ -51,7 +51,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .add_plugins(DefaultPlugins)
         .add_systems(Startup, setup_system)
         // these systems will be executed as part of the advance frame update
-        .add_systems(GgrsSchedule, (move_cube_system, increase_frame_system))
+        .add_systems(
+            GgrsSchedule,
+            (move_cube_system, increase_frame_system).in_set(AdvanceWorldSet::Main),
```
johanhelsing, line 56: I'm not sure I like the extra boiler-plate this `AdvanceWorldSet` adds. Anything that involves time will have to explicitly belong to a set... Maybe we should do it in a separate schedule, although it leads to slightly less parallel code?

johanhelsing (commented):
It would be really nice to have `Time` just work! And nice write-up :)

Side note: I've been planning on collecting small rollback utitilities/hacks like this in https://github.com/johanhelsing/bevy_roll_safe (which has an optional dependency on bevy_ggrs), because I'd also like to experiment with client/server deterministic rollback... Maybe we should also add a rollback safe state helper to ggrs itself (and perhaps flip the dependency)

src/time.rs
bushrat011899: I do agree, we can always assume there is a `Time` resource to overwrite, since GGRS already relies on its existence! As for `clone_from`, I use `swap` here to ensure I never actually overwrite the `Time` value managed by Bevy, only replace it during the `GgrsSchedule`

examples/box_game/box_game_spectator.rs
bushrat011899: Agreed, that was a DX regression. I've added a new `AdvanceWorld` schedule (which better aligns with `SaveWorld` and `LoadWorld` naming). Now, the `run_ggrs_schedules` system will run `LoadWorld`, `AdvanceWorld`, and `SaveWorld`. The original `GgrsSchedule` is the run during the `AdvanceWorldSet::Main` set by a dumb system:

```rust
|world: &mut World| world.run_schedule(GgrsSchedule)
```

This allows for easily adding systems before and after the `GgrsSchedule` using `AdvanceWorld` and `AdvanceWorldSet`, and also allows for game logic to just be added to `GgrsSchedule` as before.

bushrat011899:
> It would be really nice to have `Time` just work!

I agree!

> And nice write-up :)

Thanks!

> Side note: I've been planning on collecting small rollback utitilities/hacks like this in [bevy_roll_safe](https://github.com/johanhelsing/bevy_roll_safe) (which has an optional dependency on bevy_ggrs), because I'd also like to experiment with client/server deterministic rollback... Maybe we should also add a rollback safe state helper to ggrs itself (and perhaps flip the dependency)

I think it would be nice to build into `bevy_ggrs` as many helpers as reasonable to make the interaction between GGRS and Bevy as seamless as possible. In an ideal world, an end user could add the GGRS plugin, and rollback networking "just works". Obviously there's no one-size-fits-all solution though, which is why I'm leaning towards `bevy_ggrs` being a collection of plugins that each add a small but meaningful piece to the overall rollback experience.

I think that direction should not just give more freedom to the end users of `bevy_ggrs`, but also makes development more approachable, since the architecture is transparently written into Bevy through plugins, systems, resources, etc.

johanhelsing:
Perhaps not in this PR, but I think we should consider running some of the schedules single-threaded, as the docs say "facilitator" schedules don't benefit from multi-threading iirc.

src/time.rs
```diff
@@ -0,0 +1,117 @@
...
+    AdvanceWorld, AdvanceWorldSet, CloneStrategy, ResourceSnapshotPlugin, RollbackFrameCount,
+    DEFAULT_FPS,
+};
+
+/// [`Resource`] describing the rate at which the [`AdvanceWorld`] will run.
+///
+/// Updating this value after the session has started may produce
+/// unintended results and should be avoided.
+#[derive(Resource, Clone, Copy, Debug, Hash)]
+pub struct RollbackFrameRate(pub usize);
```
johanhelsing, line 18:
```suggestion
/// [`Resource`] describing the rate at which the [`AdvanceWorld`] will run.
#[derive(Resource, Clone, Copy, Debug, Hash, Deref)]
pub struct RollbackFrameRate(usize);
```

Maybe we should just make it impossible to modify or construct outside the crate?

bushrat011899:
> Perhaps not in this PR, but I think we should consider running some of the schedules single-threaded, as the docs say "facilitator" schedules don't benefit from multi-threading iirc.

Only on WASM, for every other platform, schedules will run multi-threaded [by default](https://docs.rs/bevy_ecs/latest/bevy_ecs/schedule/enum.ExecutorKind.html). However, I do still agree how the schedules are run is something worth revisiting, as I suspect more of the `run_schedules` system could be split out into smaller (easier to debug) systems.

src/time.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+/// A [`Resource`] which manages a [`Time`] directly tied to the [`GgrsFrameRate`].
+///
+/// This resource is available at all times, and will be updated during [`AdvanceWorld`].
+/// During this schedule, the Bevy managed [`Time`] will be replaced with one from
+/// this resource.
+#[derive(Resource, Clone)]
+pub struct RollbackTime {
```
johanhelsing, line 29: Should this be `pub(crate)`? Is there a reason for the user to interact with this instead of `Time`?

src/time.rs
```diff
@@ -0,0 +1,117 @@
...
+        let runtime = Duration::from_nanos(runtime_nanos);
+
+        // Absolute formula avoids accumulating error.
+        let now = ggrs_time.start + runtime;
+
+        ggrs_time.time.update_with_instant(now);
+    }
+
+    /// Insert a [`Time`] from the [`RollbackTime`] managed by this plugin.
+    pub fn insert(mut ggrs_time: ResMut<RollbackTime>, mut bevy_time: ResMut<Time>) {
```
johanhelsing, line 84: This should probably be updated/renamed now? To better reflects that it replaces bevy time? apply/restore perhaps?

src/time.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut time = ggrs_time.time.clone();
+
+        std::mem::swap(bevy_time.as_mut(), &mut time);
+
+        ggrs_time.bevy_time = Some(time);
+    }
+
+    /// Replace the original [`Time`] resource if it exists.
+    pub fn remove(mut ggrs_time: ResMut<RollbackTime>, mut bevy_time: ResMut<Time>) {
```
johanhelsing, line 93:
```suggestion
    /// Restore the [`Time`] resource to its original state.
    pub fn restore(mut ggrs_time: ResMut<RollbackTime>, mut bevy_time: ResMut<Time>) {
```

gschup:
Useful PR!

Is there a benefit of replacing time and swapping it back after the schedule over just providing an additional `RollbackTime` that users can access in addition to the regular `Time`?

johanhelsing (commented):
Some minor comments, but I think this is almost good to go :)

johanhelsing:
> Useful PR!
> 
> Is there a benefit of replacing time and swapping it back after the schedule over just providing an additional `RollbackTime` that users can access in addition to the regular `Time`?

One nice thing is that it's possible to use with 3rd-party plugins that don't know what `bevy_ggrs` is. For instance bevy_xpbd allows being added to a custom schedule (it also has manual stepping, so not technically needed there, though).

gschup:
> I think it would be nice to build into `bevy_ggrs` as many helpers as reasonable to make the interaction between GGRS and Bevy as seamless as possible. In an ideal world, an end user could add the GGRS plugin, and rollback networking "just works". Obviously there's no one-size-fits-all solution though, which is why I'm leaning towards `bevy_ggrs` being a collection of plugins that each add a small but meaningful piece to the overall rollback experience.

The problem I see is that we will probably never cover all features that bevy provides (or we will arrive in maintenance hell). As long as we keep these "feature fixes" in separate opt-in plugins, I like both the approach of having them in this crate or providing a separate crate. We might even consider having them as crate features, although I am not sure of the benefits.

johanhelsing:
> > Perhaps not in this PR, but I think we should consider running some of the schedules single-threaded, as the docs say "facilitator" schedules don't benefit from multi-threading iirc.
> 
> Only on WASM, for every other platform, schedules will run multi-threaded [by default](https://docs.rs/bevy_ecs/latest/bevy_ecs/schedule/enum.ExecutorKind.html). However, I do still agree how the schedules are run is something worth revisiting, as I suspect more of the `run_schedules` system could be split out into smaller (easier to debug) systems.

I thought it was that the overhead from starting a multithreaded schedule is bigger than the benefit from parallelizing it? ref this comment: https://github.com/bevyengine/bevy/blob/d3e41e2ff7653bd726854cd0a36e146e517a6b7c/crates/bevy_app/src/main_schedule.rs#L163

bushrat011899:
> Useful PR!

Thanks!

> Is there a benefit of replacing time and swapping it back after the schedule over just providing an additional `RollbackTime` that users can access in addition to the regular `Time`?

My rational is three-fold:

1. Accessing the Bevy `Time` within the `GgrsSchedule` is (probably) wrong, since any value gotten from it will be completely different during rollback. By replacing it, it is inaccessible during the schedule, protecting against that mistake (again, a mistake in my opinion)
2. Using the Bevy `Time` resource instead of a custom `RollbackTime` opens up the possibility of third-party time-dependent systems within the rollback schedule working as-is for not really any added cost, since a resource needs to be added, it might as well be the Bevy one.
3. Users will likely already be familiar with the existing `Time` resource, so a consistent type avoids any friction in transferring skills. Or in a world where we replicate the `Time` functionality in `RollbackTime`, it would avoid the effort in keeping `RollbackTime` up-to-date with `Time`'s APIs.

bushrat011899:
> The problem I see is that we will probably never cover all features that bevy provides (or we will arrive in maintenance hell). As long as we keep these "feature fixes" in separate opt-in plugins, I like both the approach of having them in this crate or providing a separate crate.

Completely agree! When I say ideal, I mean in an (unfortunately) unreachable sense. As in, if we _could_ have full feature parity with Bevy, that would be perfect. I think what is _practical_ is to provide fixes (optional or not) for the most common pain points (`Entity` rollback, `Parent`/`Child` rollback, `Time`, etc.).

> We might even consider having them as crate features, although I am not sure of the benefits.

The only benefit I can think of is slightly faster compilation time, but I don't think that would manifest unless we had a _lot_ of extra plugins!

bushrat011899:
> I thought it was that the overhead from starting a multithreaded schedule is bigger than the benefit from parallelizing it, even on multi-threaded platforms? ref [this comment](https://github.com/bevyengine/bevy/blob/d3e41e2ff7653bd726854cd0a36e146e517a6b7c/crates/bevy_app/src/main_schedule.rs#L163)

That is a good point, but I don't know if we can make that decision on behalf of someone else's game, since the schedules are all public facing, allowing the end user to add anything to them they see fit. Regardless, I don't think it even matters, since a user can take advantage of that public API to control that behaviour themselves:

```rust
// Run LoadWorld in a single thread, and also clear deferred immediately.
app.edit_schedule(LoadWorld, |schedule| {
    schedule.set_executor_kind(bevy::ecs::schedule::ExecutorKind::Simple);
});

// Run SaveWorld in multithreaded
app.edit_schedule(SaveWorld, |schedule| {
    schedule.set_executor_kind(bevy::ecs::schedule::ExecutorKind::MultiThreaded);
});
```

src/time.rs
bushrat011899: I left this public just in case an end-user wanted to get access to the game-time from outside the rollback schedules. It could be private to the crate to reduce the API surface tho.

johanhelsing:
This is Bevy's documentation for `fixed_timestep`
> ```
> //! When using fixed time steps, it is advised not to rely on [`Time::delta`] or any of it's
> //! variants for game simulation, but rather use the value of [`FixedTime`] instead.
> ```
So they don't tamper with `Time`, just warns against using it... Still, it feels kind of nice that it's just not possible to make that mistake.

An alternative would be to use a resource_scope to pull the `Time` resource out before running the `AdvanceWorld` schedule using a resource scope: https://docs.rs/bevy_ecs/latest/bevy_ecs/world/struct.World.html#method.resource_scope That would make it an error to use Time within the schedule, but not have the benefit of 3rd party plugins just working, though.

src/snapshot/set.rs
```diff
@@ -67,7 +74,32 @@ impl Plugin for SnapshotSetPlugin {
...
+        )
+        .add_systems(
+            AdvanceWorld,
+            apply_deferred
+                .after(AdvanceWorldSet::Main)
+                .before(AdvanceWorldSet::Last),
+        )
+        .add_systems(
+            AdvanceWorld,
+            (|world: &mut World| world.run_schedule(GgrsSchedule)).in_set(AdvanceWorldSet::Main),
```
johanhelsing, line 102: Since `AdvanceWorld` is linear and mostly just runs `GgrsSchedule`, and not really intended for the user to modify (?) I think its executor should be set to single-threaded. The user could always change it back if they do something weird.

The user can change this, but I think it's best if we provide the most sensible defaults.

src/time.rs
johanhelsing: I think it depends on whether we encourage the user to use `Time` or not (is it a mistake or the intended way?).

As it reads now, box_game uses `Time`, so I think it would be best to hide as an implementation detail.

examples/box_game/box_game.rs
```diff
@@ -161,31 +161,34 @@ pub fn increase_frame_system(mut frame_count: ResMut<FrameCount>) {
 pub fn move_cube_system(
     mut query: Query<(&mut Transform, &mut Velocity, &Player), With<Rollback>>,
     inputs: Res<PlayerInputs<BoxConfig>>,
+    time: Res<Time>,
```
johanhelsing, line 164: I think we should either use `RollbackTime` here, or add a comment explaining that `Time` is safe to use.

bushrat011899: I've also added some documentation to `GgrsPlugin` explaining what parts of the ECS are being rollback managed by default.

src/lib.rs
```diff
@@ -176,20 +188,30 @@ impl<C: Config> Default for GgrsPlugin<C> {
...
+            .init_schedule(ReadInputs)
+            .init_schedule(LoadWorld)
+            .init_schedule(SaveWorld)
+            .add_schedule(AdvanceWorld, {
+                let mut schedule = Schedule::default();
+                // AdvanceWorld is mostly a facilitator for GgrsSchedule, so SingleThreaded avoids overhead
+                // This can be overridden if desired.
+                schedule.set_executor_kind(ExecutorKind::SingleThreaded);
+                schedule
+            })
```
johanhelsing, line 206: Possibly scope creep: but we probably want to enable ambiguity detection (error) for these as well?

src/snapshot/set.rs
```diff
@@ -67,7 +74,32 @@ impl Plugin for SnapshotSetPlugin {
...
+            apply_deferred
+                .after(AdvanceWorldSet::First)
+                .before(AdvanceWorldSet::Main),
+        )
+        .add_systems(
+            AdvanceWorld,
+            apply_deferred
+                .after(AdvanceWorldSet::Main)
+                .before(AdvanceWorldSet::Last),
+        )
```
johanhelsing, line 99: Honest question: What's the practical difference between using `Schedule`s and sets, when they have hard sync points like apply_deferred between them?

nezuo:
I ran into non-determinism issues in the past because I was using a `bevy_xpbd` setting that relies on `Time`. This would fix that issue, so this is great!

I want to bring up the changes to the `Time` resource that are [coming in bevy 0.12](https://github.com/bevyengine/bevy-website/blob/42e42c09c91a6306cf0fb26f7c9c0394061c6156/content/news/2023-10-21-bevy-0.12/index.md#unified-time). I think we should hold off on this until that releases. A potential approach is to add a `Time<Rollback>` resource and replace `Time<()>` with it.

> Accessing the Bevy Time within the GgrsSchedule is (probably) wrong, since any value gotten from it will be completely different during rollback. By replacing it, it is inaccessible during the schedule, protecting against that mistake (again, a mistake in my opinion)

With 0.12, it will be possible to manually use `Time<Virtual>`. I agree that it's likely a mistake but there is an option to use it.

johanhelsing:
Didn't realize 0.12 was so close. And agree about using Time<Rollback>. Seems like a neat solution to the problem.

bushrat011899:
Agreed. I was aware there were some major changes to `Time` in 0.12 coming soon, but wasn't certain how soon. I'll mark this PR as draft for now, and I'm happy to wait and see exactly what gets shipped in 0.12 before committing to this.

src/snapshot/set.rs
bushrat011899: The differences (as I understand it):

- Sets must be run completely, and must be contained within a schedule. It is not possible to run all systems within `MySet::A` and not `MySet::B`.
- Systems can be ordered `before` and `after` sets, but this cannot be done for schedules.

So if we had `GgrsSchedule` on its own, there wouldn't be a simple way to schedule systems to execute `before` and `after` it runs. Adding a set, `AdvanceWorldSet` allows for controlled ordering, but then every system would need to be added in a explicit set if `AdvanceWorldSet` was inside `GgrsSchedule`. Conversely, if `GgrsSchedule` was within the `AdvanceWorldSet`, we'd have no way to trigger the schedule on demand. So we require a `Schedule` -> `Set` -> `Schedule` architecture to allow for `before` and `after` ordering, _and_ the ability to trigger the schedule whenever required.

bushrat011899:
I've updated this PR to Bevy 0.12 based on #83, which drastically simplifies the process of managing a custom `Time` resource.

gschup:
Before I merge this, would a rebase from main help to clean up the commit history of this branch?

bushrat011899:
> Before I merge this, would a rebase from main help to clean up the commit history of this branch?

Completely agree! I'm still quite new to tools like rebase, so I hope I've done this correctly. Please feel free to tell me otherwise!

gschup:
With git, there is always more to learn. Thanks for doing the work on this PR!

src/time.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub fn replace_default_with_ggrs(
+        mut default_time: ResMut<Time<()>>,
+        ggrs_time: Res<Time<GGRSTime>>,
+    ) {
+        *default_time = ggrs_time.as_generic();
+    }
+
+    pub fn replace_default_with_virtual(
+        mut default_time: ResMut<Time<()>>,
+        virtual_time: Res<Time<Virtual>>,
```
johanhelsing, line 52: Maybe a bit theoretical, but can we be completely sure Virtual if the right variant to restore? Maybe not worth it to restore the exact time we overwrite, though...

src/time.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct RollbackFrameRate(pub(crate) usize);
+
+impl Default for RollbackFrameRate {
+    fn default() -> Self {
+        Self(DEFAULT_FPS)
+    }
+}
+
+#[derive(Default, Clone, Copy, Debug)]
+pub struct GGRSTime;
```
johanhelsing, line 21: We use lower case acronyms in the rest of this crate

johanhelsing: Also needs docs

src/time.rs
bushrat011899: My rationale here is that we add these schedules to the `MainSchedule` with the `GgrsPlugin`, so `Virtual` is the correct one in the default case. However, you are correct that if instead we used the `FixedUpdate` schedule this would be wrong.

johanhelsing: Alright, I'm probably worrying too much :)

bushrat011899:
> With git, there is always more to learn. I don't think I will ever feel 100% confident with git. Thanks for doing the work on this PR!

No problems at all, I've been meaning to contribute more to the projects I use, and this has been a great project to work on, thanks!

johanhelsing (approved):
Looks great! Looking forward to use it!

src/lib.rs
```diff
@@ -132,8 +131,30 @@ pub struct LoadWorld;
...
+///
+/// # Rollback
+///
+/// This will provide rollback management for the following items in the Bevy ECS:
+/// - [Entities](`Entity`)
+/// - [Parent] and [Children] components
+/// - [Time]
+///
+/// To add more data to the rollback management, see the methods provided by [GgrsApp].
+///
```
nezuo, line 157: This comment is repeated twice.

bushrat011899: That looks like a casualty of my rebase, I'll fix that in a separate PR now. Thanks for spotting it!

PR #82, opened 2023-10-31
Ensure `Rollback` Order is Stable After Rollback

bushrat011899:
# Objective

If a `Rollback` entity is spawned, then rollback causes that entity to no longer be spawned (or vice versa), the generation of that `Entity` will be different across clients that did/not rollback. Because `Entity` derives its `PartialOrd`, the generation value changes the order of the entities, and likewise for `Rollback`. Any code relying on a stable ordering of `Rollback` components will then likely desync.

This should not happen!

# Solution

- I've removed the `PartialOrd` and `Ord` implementations for `Rollback`, since they are _not_ stable across clients, and therefore using them is a footgun waiting to happen.
- Simplified `RollbackOrdered::push` to simply stack the values in the order `AddRollbackCommand` is called. I believe this ordering _is_ stable across clients so it can be used.
- Implemented `Clone` for `RollbackOrdered` and added it to the rollback schedule to account for `Rollback` entities de/spawned during rollback.
- Added a default checksum plugin that will ensure the number of active `Rollback` entities, and the total number spawned `Rollback` entities is a part of the checksum. These are cheap since both are single `len()` calls.

# Notes

I've tested rebasing #78 onto this PR and from my testing, all desync issues are resolved. At a minimum, the frequency of desync issues is so drastically reduced that I was not able to replicate them no matter how randomly and quickly I switched peers/inputs.


src/rollback.rs
```diff
@@ -8,7 +8,7 @@ use bevy::{
 ///
 /// You must use the `AddRollbackCommand` when spawning an entity to add this component. Alternatively,
 /// you can use the `add_rollback()` extension method provided by `AddRollbackCommandExtension`.
-#[derive(Component, Hash, PartialEq, Eq, Clone, Copy, Debug, PartialOrd, Ord)]
+#[derive(Component, Hash, PartialEq, Eq, Clone, Copy, Debug)]
 pub struct Rollback(Entity);
```
johanhelsing, line 12: Aaaah. I totally forgot about https://github.com/gschup/bevy_ggrs/pull/58, and thought we we're still using "rollback ids"

johanhelsing (approved):
Awesome! I'm not seeing any desyncs in #78 anymore either. Would be really nice to have a unit test for this, but also really nice to get it in quickly so main isn't broken :P

PR #83, opened 2023-11-05
Applied Changes as Required for Bevy 0.12

bushrat011899:
# Objective

Bare minimum changes required to move to Bevy 0.12 from 0.11.

# Solution

- Bumped Bevy to version 0.12
- Split `ComputedVisibility` into `ViewVisibility` and `InheritedVisibility`
- Updated `add_schedule` calls to use new convention
- Replaced `EntityMap` with `HashMap`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -176,20 +176,19 @@ impl<C: Config> Default for GgrsPlugin<C> {
 
 impl<C: Config> Plugin for GgrsPlugin<C> {
     fn build(&self, app: &mut App) {
-        let mut schedule = Schedule::default();
-        schedule.set_build_settings(ScheduleBuildSettings {
-            ambiguity_detection: LogLevel::Error,
-            ..default()
-        });
-
         app.init_resource::<RollbackFrameCount>()
             .init_resource::<ConfirmedFrameCount>()
             .init_resource::<MaxPredictionWindow>()
             .init_resource::<RollbackOrdered>()
             .init_resource::<LocalPlayers>()
             .init_resource::<FixedTimestepData>()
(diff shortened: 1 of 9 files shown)


johanhelsing:
Once this and #80 is in, we should probably do another crates release? Eager to update tutorials and other crates to all the great new stuff!

gschup:
That sounds great! Will go over the PRs over this week

src/lib.rs
```diff
@@ -176,20 +176,19 @@ impl<C: Config> Default for GgrsPlugin<C> {
...
-
         app.init_resource::<RollbackFrameCount>()
             .init_resource::<ConfirmedFrameCount>()
             .init_resource::<MaxPredictionWindow>()
             .init_resource::<RollbackOrdered>()
             .init_resource::<LocalPlayers>()
             .init_resource::<FixedTimestepData>()
-            .add_schedule(GgrsSchedule, schedule)
-            .add_schedule(ReadInputs, Schedule::new())
+            .edit_schedule(GgrsSchedule, |schedule| {
```
gschup, line 185: How is this change related to updating to bevy 0.12? I see why it is nice to have ambiguity being an error, but that is an issue separate from updating to 0.12, if I understand correctly.

gschup (approved):
Thanks for updating! Seems we got lucky this time around and the changes are minor. Just one small comment, otherwise this looks great!

johanhelsing (approved):
Looks great to me

src/lib.rs
johanhelsing: I think it was just done further up in the file before?

gschup: Oh, I see! silly me :)

PR #86, opened 2023-11-07
Support for Serde Rollback

bushrat011899:
# Objective

In my personal project, I've hit an issue where the [Rapier](https://github.com/dimforge/bevy_rapier) library has a large state object, `RapierContext`, which cannot be cloned, copied, or reflected for snapshotting. However, it does implement Serde `Serialize + Deserialize`. It would be convenient and broadly applicable if snapshots natively supported Serde.

# Solution

I've created an **optional** snapshot strategy using Serde and Postcard, gated by the newly added `serde` feature:

```rust
// With serde feature enabled
app.rollback_resource_with_postcard::<RapierContext>();
```

# Notes

I haven't done extensive performance testing with this strategy, but I think it's a safe bet it's probably the slowest. This is why I've chosen Postcard as the target format for de/serialization, as it is designed for embedded `no-std` use, where performance would be highly valued.

I also don't believe Rapier requires this addition to work, and this could also be added by an end user entirely through the current public API. However, I can imagine many others potentially reaching for this functionality, and I wrote it for myself, so I figured I might as well offer it here as well!

Finally, I've made this branch based off of my #83 for testing purposes, but these changes could be easily ported to any alternate Bevy 0.12 upgrade branch as required.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,9 +13,11 @@ categories = ["network-programming", "game-development"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["serde"]
+default = ["serde", "ron", "bincode"]
 wasm-bindgen = ["instant/wasm-bindgen", "ggrs/wasm-bindgen"]
-serde = ["dep:serde", "dep:postcard"]
+serde = ["dep:serde"]
+ron = ["serde", "dep:ron"]
+bincode = ["serde", "dep:bincode"]
 
 [dependencies]
 bevy = { version = "0.12", default-features = false }
@@ -24,15 +26,16 @@ instant = { version = "0.1", optional = true }
 log = "0.4"
 #ggrs = { version= "0.9.4", features=["sync-send"]}
 ggrs = { git = "https://github.com/gschup/ggrs", features=["sync-send"]}
-serde = { version = "1.0.130", optional = true }
-postcard = { version = "1.0.8", optional = true, features = ["use-std", "alloc"] }
+serde = { version = "1", optional = true }
+ron = { version = "0.8", optional = true }
+bincode = { version = "1.3", optional = true }
 
 [dev-dependencies]
 bevy = { version = "0.12", default-features = true }
 clap = { version = "4.4", features = ["derive"] }
 rand = "0.8.4"
 rand_xoshiro = "0.6"
-serde = "1.0.130"
+serde = "1"
 serde_json = "1.0"
 serial_test = "2.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,9 @@ categories = ["network-programming", "game-development"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
+default = ["serde"]
 wasm-bindgen = ["instant/wasm-bindgen", "ggrs/wasm-bindgen"]
+serde = ["dep:serde", "dep:postcard"]
 
 [dependencies]
 bevy = { version = "0.12", default-features = false }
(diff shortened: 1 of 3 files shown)


gschup:
I remember writing a ggrs rapier demo at some point in the past, where I had to serialize the `RapierContext`, too. Performance wise, It was not very good. I am not sure we should offer this way of saving and loading state as a part of the crate. I am worried it might mislead users to use it in all the wrong places :)

Is there a way users can add this plugin themselves without it being distributed as part of this crate? What do you think @johanhelsing ?

bushrat011899:
> I remember writing a ggrs rapier demo at some point in the past, where I had to serialize the `RapierContext`, too. Performance wise, It was not very good. I am not sure we should offer this way of saving and loading state as a part of the crate. I am worried it might mislead users to use it in all the wrong places :)
>
> Is there a way users can add this plugin themselves without it being distributed as part of this crate? What do you think @johanhelsing ?

Oh yes, that's why I was so excited to have the `SaveWorld` and `LoadWorld` schedules added, because my entire PR (ignoring the convenience methods added to `GgrsApp`) could be collapsed to:

```rust
use std::marker::PhantomData;
use postcard::{from_bytes, to_allocvec};
use serde::{de::DeserializeOwned, Serialize};
use crate::Strategy;

/// A [`Strategy`] based on [`Reflect`] and [`FromWorld`]
pub struct PostcardStrategy<T: Serialize + DeserializeOwned>(PhantomData<T>);

impl<T: Serialize + DeserializeOwned> Strategy for PostcardStrategy<T> {
    type Target = T;

    type Stored = Vec<u8>;

    #[inline(always)]
    fn store(target: &Self::Target) -> Self::Stored {
        to_allocvec(target).unwrap()
    }

    #[inline(always)]
    fn load(stored: &Self::Stored) -> Self::Target {
        from_bytes(stored).unwrap()
    }
}

// ...

app.add_plugins(ComponentSnapshotPlugin::<PostcardStrategy<RapierContext>>::default());
```

Where the only type imported is the public `Strategy`. That's part of my motivation for keeping the plugins behind `GgrsApp` public, as it allows for some pretty clean end-user customisation. If you don't believe this is a good fit for the main repo I am totally fine with leaving this as something I personally use in my project. I just thought I'd share it as a PR since I already made it.

Cargo.toml
```diff
@@ -13,7 +13,9 @@ categories = ["network-programming", "game-development"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
+default = ["serde"]
```
johanhelsing, line 16: Perhaps not enable it by default? Does bevy depend on serde or postcard?

johanhelsing (commented):
I think it could be useful to have this for cases like the rapier serialization.

Cargo.toml
bushrat011899: Good point! I've had a look through `bevy`, `matchbox` (with `ggrs` feature enabled), and `ggrs`. Out of those three crates, `serde` version 1 is always included, `ron` version 0.8 is included (from `bevy`), and `bincode` is included (`matchbox` and `ggrs`)

I've updated this PR to use `ron` and `bincode` instead, as they are already in the dependency graph. I've still kept them behind feature flags so you can remove them if desired.

cscorley:
Just wanted to call-out here that while this strategy is cool and could be very useful for integrating with Rapier directly via your own `RapierContext`, I'm not sure it's compatible with `bevy_rapier` directly.

There are several `HashMap`s that are not serialized as part of the `bevy_rapier::RapierContext` which map between Bevy `Entity` to various Rapier `*Handle` types (e.g., `entity2collider`).  Without this, saving/restoring directly to the `bevy_rapier::RapierContext` means the plugin loses its context immediately and bevy_rapier will not work for anything the App changes (user inputs to update `Transforms`, `Velocity`, etc.).  Regular, non-Bevy physics continues to work, however.

I am curious to know how you've handled this case.  For me, I've settled on creating N "blessed entities" and initializing Rapier with all of them, promising myself I can only rollback those N entities.  Later, I will do all the (de)serializing directly and then only updating the appropriate properties on the `bevy_rapier::RapierContext`.

(I have not looked into this yet, but I'm assuming those maps are not serialized because `Entity` isn't serializable.  I don't see why `Entity` couldn't be, though!)

gschup:
Thanks for this contribution! 

After leaving it open for a long, long time, I've decided not to merge this into the core crate for two reasons:

- The serde strategy doesn't fully cover the bevy_rapier case due to unmapped internal HashMaps between Bevy entities and Rapier handles (as noted by @cscorley).
- Since the Strategy trait is public and designed for exactly this kind of extension, I think users should implement these specific cases themselves.

We can always discuss and reopen if anyone sees the need!

PR #87, opened 2023-11-07
Feature: Add cargo fmt check to github action

gschup:
We already do this in the ggrs crate, I feel it should be done here, too.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,3 +22,5 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+    - name: Check formatting
+      run: cargo fmt --check
--- a/src/snapshot/entity_checksum.rs
+++ b/src/snapshot/entity_checksum.rs
@@ -24,10 +24,7 @@ impl EntityChecksumPlugin {
 
         let result = ChecksumPart(hasher.finish() as u128);
 
-        trace!(
-            "Rollback Entities have checksum {:X}",
-            result.0
-        );
+        trace!("Rollback Entities have checksum {:X}", result.0);
 
         if let Ok(mut checksum) = checksum.get_single_mut() {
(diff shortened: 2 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -22,3 +22,5 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+    - name: Check formatting
+      run: cargo fmt --check
```
johanhelsing, line 26: Here's what I've been doing in my crates:

```
      - name: Fmt
        run: cargo fmt --all -- --check
```

Haven't checked what the difference is

gschup: From `cargo fmt --help`:  
```
--all
Format all packages, and also their local path-based dependencies
```

johanhelsing: So i guess it will probably only be relevant if we want to add derive suport

src/snapshot/entity_checksum.rs
```diff
@@ -24,10 +24,7 @@ impl EntityChecksumPlugin {
 
         let result = ChecksumPart(hasher.finish() as u128);
 
-        trace!(
-            "Rollback Entities have checksum {:X}",
-            result.0
-        );
+        trace!("Rollback Entities have checksum {:X}", result.0);
```
bushrat011899, line 27: Thanks for doing this. I kept tripping this change while I was working on the other PRs and kept reverting it since it wasn't a part of my current work!

bushrat011899:
A good change! Should `clippy` be added as well? Currently it doesn't flag any issues, but it's definitely helped me during development.

gschup:
While I love working with clippy, I am not sure it is correct to fail the build if clippy has something. Some of the warnings are not always necessary and I am not the biggest fan of putting clippy ignore annotations in the code.

PR #89, opened 2023-11-08
Add respawn-entity test

johanhelsing:


examples/manual_tests/respawn_entity.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn default() -> Self {
+        Self(3)
+    }
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut session = SessionBuilder::<GgrsConfig>::new()
+        .with_num_players(1)
+        // each frame, roll back and resimulate 5 frames back in time, and compare checksums
+        .with_check_distance(2);
```
nezuo, line 41: Won't a `with_check_distance` of 2 only resimulate 5 frames?

nezuo (commented):
It might be useful to add comment to the top of the test that says it's purpose is to make sure despawning and rollback don't cause a panic.

examples/manual_tests/respawn_entity.rs
johanhelsing: It's better to increase it yes. I set it to to for a very minimal repro of the issue, but should revert that last commit.

nezuo: I just realized I mistyped my comment. I meant to say: "Won't a `with_check_distance` of 2 only resimulate 2 frames?" The comment says 5.

PR #95, opened 2023-11-18
Schedule systems after input

nezuo:
Bevy's input systems run in `PreUpdate` in the `InputSystem` set. This ensures that the `ReadInputs` schedule runs after input is updated for the frame.


johanhelsing (approved):
Nice catch!

gschup (approved):

PR #99, opened 2024-01-05
Implement update for CloneStrategy

370417:
Use `clone_from` in `CloneStrategy::update`. Could be more efficient if the user has overridden `clone_from`.


johanhelsing (approved):
Looks good to me!

gschup (approved):

Issue #98, opened 2023-12-19
Checksums are not portable


aleksa2808:
**Describe the bug**

When running `bevy_ggrs` on two different architectures (in my case these were 64bit x86 and 32bit WASM) false desyncs can occur because the checksum code is not portable. Several reasons contribute to this:
1. `bevy::utils::FixedState` is used for calculating checksums which uses the non-portable `aHash` crate, meaning that two machines can give two different checksums from the same number. Quoting the `aHash` crate [repo](https://github.com/tkaitchuck/ahash):
```
Because it does not have a fixed standard, different computers or computers on different versions of the code will observe different  
hash values. As such, aHash is not recommended for use other than in-memory maps. Specifically, aHash is not intended for network  
use or in applications which persist hashed values. (In these cases HighwayHash would be a better choice)
```
2. Checksums are sometimes calculated from `usize`, for example the return of a `.len()` call. Even with a portable hasher, from the same source `usize` number you would get different hashes on 32bit and 64bit machines, as this type is 32 bit and 64 bit wide, respectively.

I resolved these issues (or at least up to what my game needed) in a [fork branch](https://github.com/gschup/bevy_ggrs/commit/5b9509291635a470cdddce0a4b90c60f4a8cd7cd). Here, I did the following:
1. Used a forked version of a portable hasher, SeaHash. The only reason for forking was to add panics when trying to hash `usize`/`isize` variables, as a prevention measure.
2. Limited the use of `usize`/`isize` types in `bevy_ggrs`, usually by using or casting to `u64`.

**Expected behavior**

I expect the checksums to be the same across architectures.
I'm not yet sure what would be the best solution for `usize`/`isize` variables.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Opera

johanhelsing:
Merging #94 first would make fixing this a bit simpler, because then it's just one place to change the hasher.

johanhelsing:
@aleksa2808 Thanks for the detailed write-up. Do you want to contribute this as a PR (casting to u64 and replacing with SeaHash or another portable hasher)?

aleksa2808:
[...]

PR #104, opened 2024-02-14
Fix checksum portability

aleksa2808:
This PR fixes checksum portability across different architectures by:
1. Replacing the usage of `usize` with `u64` in places where these numbers were hashed.
2. Switching the hasher implementation from the Bevy-provided, non-portable `AHash` hasher to the portable `seahash` hasher. 

Fixes https://github.com/gschup/bevy_ggrs/issues/98.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,11 +17,12 @@ wasm-bindgen = ["instant/wasm-bindgen", "ggrs/wasm-bindgen"]
 
 [dependencies]
 bevy = { version = "0.12", default-features = false }
-bytemuck = { version = "1.7", features=["derive"]}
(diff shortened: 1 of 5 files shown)


gschup (approved):

johanhelsing:
Thanks! If someone wants to add a test for this, we could perhaps look into using the [`insta`](https://docs.rs/insta/latest/insta/) for snapshot testing of bevy_ggrs snapshot checksum values, and run test on two different architectures.

johanhelsing (approved):

johanhelsing:
Or just hardcode the checksum value in the test I guess, they're not that long...

PR #106, opened 2024-02-18
Update to Bevy 0.13

### LINKED ISSUE #105: Update to Bevy 0.13
Bevy 0.13 has been released, let's get this updated!

bushrat011899:
# Objective

- Fix #105

## Solution

- Bumped Bevy version to 0.13
- Updated `Input` references to `ButtonInput`
- Updated `KeyCode` values to prepend `Key` (e.g., `KeyCode::W` -> `KeyCode::KeyW`)
- Replaced `Cube` and `Plane` meshes with their new math primitives as recommended for depreciation reasons
- Explicitly converted `Color` into a `StandardMaterial` rather than relying on implicit `into()`
- Updated `AddRollbackCommandExtension` to match new lifetime generics for `EntityCommands`
- Implemented `EntityMapper` (which is now a trait instead of a type) for `RollbackEntityMap`, which will return the provided entity if it is not contained in the mapping. This ensures entities which are not tracked by the rollback map are not modified.
- Replaced the inner `HashMap` of the `RollbackEntityMap` with an `EntityHashMap`, which is an optimised type provided by Bevy.
- Updated the `Component` and `Resource` entity mapper plugins to use the new `EntityMapper` implementation provided by `RollbackEntityMap`. This removes a significant *TODO* on their implementations where if a mapping was done over a non-rollback entity, the mapping had to be undone.
- Ran `fmt` and `clippy` and ensured clean output.

Tests and the synctest example work as expected. I haven't had a chance to verify this against a real multiplayer application, so if anyone has the free time to test this branch out, please do!

### DIFF
--- a/examples/box_game/box_game.rs
+++ b/examples/box_game/box_game.rs
@@ -1,4 +1,4 @@
-use bevy::{prelude::*, utils::HashMap};
+use bevy::{prelude::*, render::mesh::PlaneMeshBuilder, utils::HashMap};
 use bevy_ggrs::{
     AddRollbackCommandExtension, GgrsConfig, LocalInputs, LocalPlayers, PlayerInputs, Rollback,
     Session,
@@ -54,7 +54,7 @@ pub struct FrameCount {
 /// Collects player inputs during [`ReadInputs`](`bevy_ggrs::ReadInputs`) and creates a [`LocalInputs`] resource.
 pub fn read_local_inputs(
     mut commands: Commands,
-    keyboard_input: Res<Input<KeyCode>>,
+    keyboard_input: Res<ButtonInput<KeyCode>>,
     local_players: Res<LocalPlayers>,
 ) {
     let mut local_inputs = HashMap::new();
@@ -62,16 +62,16 @@ pub fn read_local_inputs(
     for handle in &local_players.0 {
         let mut input: u8 = 0;
 
-        if keyboard_input.pressed(KeyCode::W) {
+        if keyboard_input.pressed(KeyCode::KeyW) {
             input |= INPUT_UP;
         }
-        if keyboard_input.pressed(KeyCode::A) {
+        if keyboard_input.pressed(KeyCode::KeyA) {
             input |= INPUT_LEFT;
         }
-        if keyboard_input.pressed(KeyCode::S) {
+        if keyboard_input.pressed(KeyCode::KeyS) {
             input |= INPUT_DOWN;
         }
-        if keyboard_input.pressed(KeyCode::D) {
+        if keyboard_input.pressed(KeyCode::KeyD) {
             input |= INPUT_RIGHT;
         }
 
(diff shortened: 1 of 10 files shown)


examples/box_game/box_game.rs
```diff
@@ -123,7 +123,7 @@ pub fn setup_system(
                 // ...add visual information...
                 PbrBundle {
                     mesh: mesh.clone(),
-                    material: materials.add(color.into()),
+                    material: materials.add(StandardMaterial::from(color)),
```
johanhelsing, line 126: Is this just a style thing?

Personally, I find it a bit verbose.

src/snapshot/component_map.rs
```diff
@@ -72,55 +69,16 @@ fn apply_rollback_map_to_component_inner<C>(world: &mut World, map: Mut<Rollback
...
 
     trace!(
         "Mapped {}",
         bevy::utils::get_short_name(std::any::type_name::<C>())
     );
-
-    // If the entity map is now larger than the set of rollback entities, then dead entities were created.
-    // TODO: This workaround is required because the current behavior of `map_all_entities` is to change all entities,
-    // creating dead entities instead of leaving them with their original value. If `EntityMapper` behavior changes,
-    // then this workaround may no longer be required.
```
johanhelsing, line 87: Begone! :D

johanhelsing (approved):
Looks great to me!

Nice to see so many red lines 🧹 

I haven't tested in any of my projects. Still lagging a bit behind and haven't even finished the 0.12 migration for cargo space.

I have enough time, I'll check with *extreme bevy* tomorrow. Otherwise, I think we should just go ahead. If anything shows up later we could always do a patch release.

src/snapshot/component_map.rs
gschup: This is my favourite part.

examples/box_game/box_game.rs
bushrat011899: Yeah I had an issue where it was complaining it didn't know what particular `Into` implementation to use. I threw this in just to get around it without too much investigation.

src/snapshot/component_map.rs
bushrat011899: Yeah it felt _really_ good to make an update PR that _removed_ more code than it added, and got us a little extra performance as well! (I assume, haven't measured, and it's probably not much anyway)

examples/box_game/box_game.rs
gschup: I don't have a strong opinion, but I am leaning towards liking it, since it is explicit about the magic bevy type that goes into this parameter.

gschup (approved):

johanhelsing:
Tested with matchbox and extreme_bevy now. Everything works like a charm :)

johanhelsing:
Okay, we just need to fix the conflict, then we can merge this :)

bushrat011899:
Done! Should be ready to go

PR #120, opened 2025-04-25
Upgrade to bevy 0.16

johanhelsing:
Update to Bevy 0.16. There were a few non-trivial issues when porting this time:

## `Parent` (now `ChildOf`) - `Children` relations

Bevy's hierarchies have been refactored. They no longer use `MapEntities` but the brand new ECS relations feature, and they are immutable components. This means our old approach to syncing hierarchies doesn't work anymore.

I wrote a new specialized plugin that still uses our mapping table, it's just used in order to create brand new `ChildOf` component when children are restored instead of mutating an existing one.

If there are other classes of components like this, user-created or otherwise, we should probably look into making a generic shared solution for components that are immutable but still require entity mapping.

## Immutable components are a challenge for ergonomics/performance

Bevy 0.16 introduced immutable components, they can't be rolled back (without violating unsafe and even then guarantees in bevy) by assigning to a `&mut` reference, which is what our snapshot `Strategy` does now.

As far as I've reasoned, the most correct thing to do with immutable components is to create and assign a new component that replaces the old one. As long as components implement `Clone` or `Copy`, this is possible to and we can even check inside the generic plugin whether we're a mutable or immutable component. The problem is that we are not allowed by the type system to create a `Query<&mut SomeComponent>` if the target component is immutable, so we can't do anything with the information if we make the snapshotting generic over the `Component` trait.

There is an option to use a `Query<EntityMut, With<SomeComponent>` instead, however performance seems to suffer quite a bit from this. One simple benchmark I did showed ~40% slowdown.

Another option is to simply duplicate the entire component snapshot plugin, and create one mutable and one immutable version. This means the user will now have to use a separate snapshotting api for immutable components:

```rust
app.rollback_immutable_component_with_copy::<SomeImmutableComponent>();
app.rollback_component_with_copy::<SomeComponent>()
```

I think this trade-off in ergonomics is worth the extra performance, considering immutable components still seem to be quite rare. Unless someone knows of a better solution?

## Refactoring plugin to better support testing and benchmarking

The previously mentioned challenges also meant that I was quite unsure about whether I'd done my porting **correctly** and also how **performant** it was. We had some tests for hierarchies and entity mapping, but they were more integration level tests, and it was a bit unclear to me exactly what they were testing.

The stress test is also nice, but not so easy to compare performance before and after a change.

For this reason, I refactored the plugin so that world snapshotting is pretty much decoupled from anything ggrs related. This means we can easily add a `SnapshotPlugin` and step, rollback and save manually in tests, making it much simpler to write assertions about what exactly is expected to happen during each of the three.

## Micro-benchmarking for saving and loading snapshots

Micro benchmarks are equally easy to write now, and a couple have been added (they are the rationale behind my decision to split component snapshotting API on mutability).

This means an added dev-dependency on criterion, I hope that's ok.

I'm also quite excited about other opportunities a standalone snapshotting plugin enables :)

***Old PR description below***

---

~~Got it to a state where it compiles and runs the box game p2p and synctest example without issues~~

Everything should work now, just some parts of the implementation are a bit messy.

- [x] Hierarchies have been heavily refactored in Bevy, and I don't think our approach with `MapEntities` will work. I've just disabled it for now. But we should probably try to see if it's possible
- [x] Check that the MapEntities code is still sane, or perhaps remove it.
- [x] Clean up deprecated API usages
  - [x] stop using `apply_deferred`
- [x] all tests should pass 
- [x] I've just added a trait bound for all components to be mutable. Maybe we need to also handle immutable components? Needs some thought.
- [ ] Maybe there are new features in Bevy 0.16 that can make things easier or more performant? Very rough ideas:
   - [ ] Use relations somehow?
   - [x] Rollback ids should probably be immutable components
   - [ ] There is a system for disabling entities, maybe we could use that instead of completely despawning when rolling back.

Input would be very welcome. Or PR could be adopted by someone else, it's very unpredictable for me when/if I get the time to continue work on this.


rcorre:
Thanks! I was actually looking  into this and got stuck on `MapEntities` as well. It seems  like `ChildOf`/`Children` _should_ implement that, but don't. I was wondering if that was just an oversight  (seems like the old `Parent`/`Children` components did), or if there's some larger concept I'm missing. I might take that question over to a `bevy` discussion

johanhelsing:
Ok, pretty sure hierarchies work now...

Immutable components are a bit tricky, because the current generic snapshotting plugin uses a mutable reference and overwrites using assignment.

Should probably write a specialization that always uses insert.

johanhelsing:
I added support for immutable components. I'm a little bit disappointed over how little code I was able to re-use from mutable components though.

Hard due to lack of template specialization in rust, or at least I think that's the term.

bushrat011899:
> Ready for review. I'm not super happy about all the solutions here, but they work, and I think we're approaching something releasable for bevy 0.16.

With a bit of `unsafe` you can increase the amount of shared code. The `Mutability` associated type on `Component` includes a `MUTABLE` associated constant, which lets you at run-time determine if a component `T` is mutable. You can then use `get_mut_assume_mutable` to get a mutable reference to the component. By name, you're only supposed to use this method if the component is actually mutable, but it _can_ be used to mutate immutable components too, you just need to ensure that doesn't violate any other invariant.

We use this pattern in the reflection implementation for component:

```rust
// Some method for a generic component `C`
let apply = |mut entity, reflected_component| {
    if !C::Mutability::MUTABLE {
        let name = ShortName::of::<C>();
        panic!("Cannot call `ReflectComponent::apply` on component {name}. It is immutable, and cannot modified through reflection");
    }

    // SAFETY: guard ensures `C` is a mutable component
    let mut component = unsafe { entity.get_mut_assume_mutable::<C>() }.unwrap();
    component.apply(reflected_component);
};
```

Because `Component::Mutability::MUTABLE` is an associated constant, it _should_ still specialize down well, since after substitution it'll become `if true { ... }` or `if false { ... }`, which is trivially optimised.

johanhelsing:
@gschup how do you feel about adding a bit of unsafe? Is it worth it to create a simpler unified API?

I think to be safe we should probably avoid overwriting mutating immutable components by reference, but simply use it to check whether we should use an insert command instead (which is what the current `ImmutableComponentSnapshotPlugin` does.

johanhelsing:
@bushrat011899 I tried unifying it now, and I got stuck.. It seems like `get_mut_assume_mutable` only exists on `EntityWorldMut`, and I'm unsure how to get that and if we even should?

Here's what I got so far:

```rust
impl<S> ComponentSnapshotPlugin<S>
where
    S: Strategy,
    S::Target: Component + ComponentMutability,
    S::Stored: Send + Sync + 'static,
{
    pub fn load(
        mut commands: Commands,
        mut snapshots: ResMut<GgrsComponentSnapshots<S::Target, S::Stored>>,
        frame: Res<RollbackFrameCount>,
        mut query: Query<EntityRef, With<Rollback>>,
    ) {
        let snapshot = snapshots.rollback(frame.0).get();

        for mut entity in query.iter_mut() {
            let (rollback, component) = entity.components::<(&Rollback, Option<&S::Target>)>();

            let snapshot = snapshot.get(rollback);

            match (component, snapshot) {
                (Some(mut component), Some(snapshot)) => {
                    if <S::Target as ComponentMutability>::MUTABLE {
                        unsafe {
                            // Error: get_mut_assume_mutable doesn't exist for EntityRef
                            let component = entity.get_mut_assume_mutable::<S::Target>();
                            S::update(component.as_mut(), snapshot);
                        }
                    } else {
                        commands.entity(entity).insert(S::load(snapshot));
                    }
                }
                (Some(_), None) => {
                    commands.entity(entity).remove::<S::Target>();
                }
                (None, Some(snapshot)) => {
                    commands.entity(entity).insert(S::load(snapshot));
                }
                (None, None) => {}
            }
        }

        trace!(
            "Rolled back {} {} component(s)",
            snapshot.iter().count(),
            disqualified::ShortName::of::<S::Target>()
        );
    }
}
```

bushrat011899:
I believe all you need to do here is replace `Query<EntityRef, ...>` with `Query<EntityMut, ...>`, as [that also has `get_mut_assume_mutable`](https://docs.rs/bevy_ecs/latest/bevy_ecs/world/struct.EntityMut.html#method.get_mut_assume_mutable).

johanhelsing:
Oh, embarrasing. I felt like I looked everywhere 🤦 .

Ok, I got it working, but now I'm a bit unsure about whether it's a good idea? `EntityMut` gives mutable access to all components on an entity, so this probably means we now break the parallelization of restoration of snapshots?

I pushed the code so you can look at it, but I think perhaps I'll revert it.

What do you think?

johanhelsing:
I had a go at refactoring bevy_ggrs to decouple snapshotting from anything ggrs specific. Went smoother than expected:

- https://github.com/johanhelsing/bevy_ggrs/pull/3

I can merge it into this branch unless anyone disagrees with the approach?

johanhelsing:
> Ok, I got it working, but now I'm a bit unsure about whether it's a good idea? `EntityMut` gives mutable access to all components on an entity, so this probably means we now break the parallelization of restoration of snapshots?

After refactoring, I was able to do some benchmarking. I tested with `EntityMut` and the old approach, and it is indeed significantly slower, even when there is only one component type.

image


```rust
#[derive(Component, Clone, Copy)]
struct Foo(i32);

fn advance_and_load(app: &mut App) {
    app.world_mut().run_schedule(AdvanceWorld);
    app.insert_resource(RollbackFrameCount(0));
    app.world_mut().run_schedule(LoadWorld);
}

fn increment_foos(mut foos: Query<&mut Foo>) {
    for mut foo in &mut foos {
        foo.0 += 1;
    }
}

fn foo_1000(c: &mut Criterion) {
    let mut app = App::new();
    app.add_plugins((MinimalPlugins, SnapshotPlugin));
    app.add_systems(AdvanceWorld, increment_foos);
    app.rollback_component_with_copy::<Foo>();
    app.update();
    app.world_mut()
        .run_system_once(|mut commands: Commands| {
            for i in 0..1000 {
                commands.spawn(Foo(i)).add_rollback();
            }
        })
        .unwrap();
    app.world_mut().run_schedule(SaveWorld);
    c.bench_function("advance_and_load_1000_components", |b| {
        b.iter(|| advance_and_load(&mut app))
    });
}
```

If you're okay with adding `criterion` as a dev dependency, I can add the benchmarks as well.

But the main question is: Do we want a nice ergonomic API or do we want performance? As far as I can see we can't have both. I'm rooting for just duplicating the plugin for immutable components, since I don't think they are that common.

@bushrat011899 if you have more input it would be very welcome. @gschup is pretty busy these days.

gschup:
I will take a look today! Just from reading the thread, this all sounds pretty nice :)

johanhelsing:
@gschup Thanks! I just updated the PR description to try and make it a little easier to follow :)

gschup:
> But the main question is: Do we want a nice ergonomic API or do we want performance? As far as I can see we can't have both. I'm rooting for just duplicating the plugin for immutable components, since I don't think they are that common.

Let's try to keep it performant. Maybe we can refactor this in the future once something inside bevy shifts again :)

EDIT from @gschup: I am so sorry. Instead of replying to your post, I edited it and thus destroyed it....

gschup (approved):

johanhelsing:
No worries, and thanks for reviewing!

PR #122, opened 2025-06-05
Make `RollbackFrameRate` fields public

### LINKED ISSUE #121: Make `RollbackFrameRate` fields public
**Is your feature request related to a problem? Please describe.**
Currently, the only way to change GGRS schedule FPS is with the `set_rollback_schedule_fps` world command. It doesn't allow for dynamic change of the GGRS schedule FPS, which while ensuring the crate's users don't mess up with the schedule, prevents for example, me to implement a network configuration file 

**Describe the solution you'd like**
If you make a newtype struct public, also just make its inner value public, else I guess it's just useless.

**Describe alternatives you've considered**
I don't know if making the `bevy_ggrs` API more open matches your coding philosophy, but if you don't like this solution I can still do some unsafe rust magic to bypass this issue.

**Additional context**
This issue I'm having also applies to some other elements of the crate, such as `ConfirmedFrameCount` newtype struct which also has a private inner field. You might consider making these ones more open too.

GaspardCulis:
Fixes #121

### DIFF
--- a/examples/stress_tests/particles.rs
+++ b/examples/stress_tests/particles.rs
@@ -2,7 +2,7 @@ use bevy::{
     color::palettes::css::ORANGE, math::vec3, platform::collections::HashMap, prelude::*,
     window::WindowResolution,
 };
-use bevy_ggrs::{checksum_hasher, prelude::*, LocalInputs, LocalPlayers};
+use bevy_ggrs::{LocalInputs, LocalPlayers, checksum_hasher, prelude::*};
 use clap::Parser;
 use ggrs::{DesyncDetection, UdpNonBlockingSocket};
 use rand::{Rng, SeedableRng};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,8 +25,8 @@ pub(crate) mod time;
 
 pub mod prelude {
     pub use crate::{
-        snapshot::prelude::*, AddRollbackCommandExtension, GgrsConfig, GgrsPlugin, GgrsSchedule,
-        GgrsTime, PlayerInputs, ReadInputs, Rollback, RollbackApp, RollbackFrameRate, Session,
+        AddRollbackCommandExtension, GgrsConfig, GgrsPlugin, GgrsSchedule, GgrsTime, PlayerInputs,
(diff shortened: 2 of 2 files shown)


GaspardCulis:
Maybe also consider making other public newtype struct fields public ?

johanhelsing:
I see... what would happen if `RollbackFrameRate` is changed mid-session, though?

Maybe it should have been part or derived from the the session resource instead?

If we make it fully mutable, then we should also get rid of `set_rollback_schedule_fps`

johanhelsing:
> Also thx for the amazing matchbox crate @johanhelsing ❤️ I'm using it for my [robot-rumble](https://github.com/GaspardCulis/robot-rumble-rs) game it's really handy.

Thanks ❤️ 

> I mean sure you could mess up the session by mutating `RollbackFramerate`; but so can you by idk changing `RollbackFrameCount` which is publicly mutable. If a dev mutates `RollbackFramerate` he's probably thinking to himself "well that might break everything if my session is started", so safeguarding it is kinda overkill to me.

Hmm... yeah, just feels nice to have an API that's impossible to misuse, if possible.

> I mean set_rollback_schedule_fps is a handy way of setting it so I think we should keep it.

My thinking is just that it's strange to have an extension method to set something so trivial. I don't think

```rust
app.set_rollback_schedule_fps(60)
```

is significantly clearer or more ergonomic than simply:

```rust
app.insert_resource(RollbackFrameRate(60);
```

And the latter is more in line with how most other bevy plugins do their config.

@gschup any opinions here?

johanhelsing:
Yes, just remove the extension method, and add a little doc comment to the resource, explaining what it does

johanhelsing (approved):
Looks good to me :)

johanhelsing:
Why do you prefix all your commits with chore? Only the last seems to be what i would personally consider a chore

GaspardCulis:
True lol, trying to follow Angular's conventional commit rules but sometimes I hesitate. I guesss I could've used the `refactor` prefix

johanhelsing:
I understand. I think most of them are actually `feat` and `feat!`, since they are changing behavior, but not fixing bugs. 

Thanks for the PR, and good luck with the configurable frame rate :)

PR #124, opened 2025-10-02
Upgrade to Bevy 0.17

johanhelsing:
Just a simple port.

- `Event -> Message`
- `WindowResolution::new` now takes `u32` instead of `f32`.
- `InputSystem` -> `InputSystems`


johanhelsing:
I suspect @gschup may still be a bit busy with life, so if anybody else has the time for a review, it would probably be very welcome :)

@rcorre, @bushrat011899?

PR #127, opened 2026-01-22
WIP Despawn using a disabling component

adammitchelldev:
This PR implements a new way of despawning entities using a disabling component that records which frame the entity was despawned on. If the simulation performs a rollback then we can simply remove the component to resurrect the entity. If instead, the confirmed frame count moves past the despawn marker, then we can safely full despawn the entity.

This is a first pass at an implemention to support further discussion, with a couple of intended features missing. It has not been tested yet.

I haven't yet looked into which other systems are going to need to build in explicit bypass for the disabling component, but my theory is that existing systems will continue to work fine. Because we resurrect the entities at the start of `LoadWorld`, the existing mechanism for respawning entities that are present in the snapshot but don't exist in the world should simply see that the resurrected entities are there, as though they never despawned.

Looking for comments/feedback on the approach and any steer on what can be accepted as a PR down the line.

### DIFF
--- a/src/snapshot/despawn.rs
+++ b/src/snapshot/despawn.rs
@@ -35,13 +35,10 @@
 use crate::snapshot::despawn::private::RollbackDespawnCommandExtensionSeal;
 use crate::{
     AdvanceWorld, AdvanceWorldSystems, ConfirmedFrameCount, LoadWorld, LoadWorldSystems,
-    RollbackFrameCount, SaveWorld, SaveWorldSystems,
+    RollbackFrameCount,
 };
 use bevy::app::{App, Plugin};
-use bevy::prelude::{
-    Children, Component, Entity, EntityCommands, EntityMut, EntityRef, EntityWorldMut,
-    IntoScheduleConfigs, Local, Query, QueryState, Res, World,
-};
+use bevy::prelude::*;
 use ggrs::Frame;
 use std::cmp::Ordering;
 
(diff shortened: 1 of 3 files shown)


adammitchelldev:
> @adammitchelldev Where is the downstream issue (if there is one)? I'm really interested in following progress on this.
> 
> Also I like the sound of a Disable approach. The despawn/respawn that currently happens is a huge footgun.

I ended up shelving my project for now, it was fun and I did spec up a kind of algorithm for doing the enable/disable and ID management, but Bevy shifts a lot and doesn't make it all that easy to mess with low level stuff. I also reckoned it might still not work even if I made progress due to new features of Avian not being designed with rollback in mind.

It's hard to see how it's realistic to achieve out of the box deterministic rollback physics without either building it into Avian or being able to exactly snapshot and restore the gamestate. My last thoughts on the matter were that the only sensible approach was going to be to create a separate Bevy World for the rollback simulation and do a pure memory snapshot of this world (I think flecs has this feature.) Ideally, you'd also make IDs deterministically synchronized between clients.

The ID allocation algorithm I came up with was quite clever but difficult to reason about.

Basically, it works like a queue of freed IDs, but with rollback support. This is achieved by not dropping the elements at the head of the queue when they are polled (i.e. a freed ID is allocated). Instead, you leave them there, and you snapshot the positions of the head and tail of the queue, so that you can rollback on a future frame. If you do a rollback, then you restore the positions from the snapshot. You only drop IDs from the queue once you can no longer roll back to that position due to the snapshots. You also don't drop the elements at the tail of the queue after rollback either, and you don't hand IDs back to Bevy at all. This way, you keep all entity IDs under your total control and ensure they always get allocated in the same order they would have been on a fresh frame (rather than expecting Bevy to do that for you.)

PR #129, opened 2026-02-22
Rollback marker

johanhelsing:
This effectively renames the old Rollback to RollbackId, and creates a
new unit struct `Rollback`

So the new flow is:

1. Insert `Rollback` on an entity in any way you want (required
   components, scenes, spawning.
2. The `OnAdd` hook handles creating a `RollbackId(Entity)` which is
   what we use internally for ordering entities.

This had the added overhead of having two components per rollback
entity. So a redundant entity list basically. But it allows much better
ergonomics, and scene authoring workflows, scripting interations++.

Before:

```rust
#[derive(Component)
struct Player;

// Have to remember to explicitly add rollback anywhere Player is spawned
commands.spawn((Player, Sprite {}).add_rollback();
```

Now:

```rust
#[derive(Component)
#[require(Rollback)
struct Player;

// Rollback added automatically :)
commands.spawn((Player, Sprite {});
```

Or simply:

```rust
app.register_required_components::<RigidBody, Rollback>();
```

or

```rust
app.require_rollback::<RigidBody>(); 
```

Then all movable physics objects will be rolled back. Don't have to remember to add rollback anywhere.

If you prefer to be explicit about it, you could still do:

```rust
commands.spawn((Player, Rollback, Sprite {}));
```

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -53,7 +53,7 @@ fn foo_1000(c: &mut Criterion) {
     app.world_mut()
         .run_system_once(|mut commands: Commands| {
             for i in 0..1000 {
-                commands.spawn(Foo(i)).add_rollback();
+                commands.spawn((Foo(i), Rollback));
             }
         })
         .unwrap();
@@ -79,9 +79,9 @@ fn foo_bar_baz_1000(c: &mut Criterion) {
     app.world_mut()
         .run_system_once(|mut commands: Commands| {
             for i in 0..1000 {
-                commands.spawn(Foo(i)).add_rollback();
-                commands.spawn(Bar(i)).add_rollback();
-                commands.spawn(Baz(i)).add_rollback();
+                commands.spawn((Foo(i), Rollback));
+                commands.spawn((Bar(i), Rollback));
+                commands.spawn((Baz(i), Rollback));
             }
         })
         .unwrap();
--- a/examples/box_game/box_game.rs
+++ b/examples/box_game/box_game.rs
@@ -1,8 +1,5 @@
 use bevy::{platform::collections::HashMap, prelude::*};
-use bevy_ggrs::{
-    AddRollbackCommandExtension, GgrsConfig, LocalInputs, LocalPlayers, PlayerInputs, Rollback,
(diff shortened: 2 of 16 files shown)


johanhelsing:
This is untested in anything but the tests in this repo, but I'd like some feedback. I think it improves ergonomics significantly, and in my case the performance tradeoff and slightly uglier internals is definitely worth it.

johanhelsing:
Would also be possible to remove the `Rollback` component and have the following API:

```rust
app.rollback_entities_with::<Player>()
   .rollback_component_with_copy::<Player>()
   .rollback_component_with_copy::<Transform>()
```

It would register an `OnAdd<Player>` hook that does the same thing. If an entity has multiple rollback marker components, they would redundantly try to add rollback for the entity, wasteful, but not sure if that's really a problem though.

johanhelsing:
... or we could potentially support the following API:

```rust
#[derive(Default, Reflect, Component, Clone, Copy, Deref, DerefMut, Rollback)]
#[rollback(copy, marker)]
#[reflect(Rollback)]
pub struct Velocity(pub Vec3);

#[derive(Resource, Default, Reflect, Hash, Clone, Copy, Rollback)]
#[rollback(resource, copy, checksum)]
#[reflect(Hash, Rollback)]
pub struct FrameCount {
    pub frame: u32,
}
```

This would auto-detect through reflect, and require no manual registration through app command extensions.

What I like about this is that everything about a type is in one place. It's harder to forget rollback registration when creating a new component alongside other components that already have it.

If reflect auto-reflect registration is disabled, the following would still work:

```rust
app.rollback::<Velocity>();
app.rollback::<FrameCount>();
app.rollback_component_with_copy::<Transform>();
```

Since we control the underlying trait, we probably implement `RollbackRegistration` for common built-in bevy types types, and get:

```rust
app.rollback::<Transform>(); // registers both hash and copy strategies without the user having to think about it.
```

Maybe it's too much magic? What do you think?

gschup (approved):
I like it! Among all proposals, I like the first one the best. the last one really is too much magic :D

PR #131, opened 2026-06-14
Fix GgrsTime panic on session restart

johanhelsing:
`update()` called `advance_to()` unconditionally, which panics when the frame count moves backwards (e.g. stopping a session and starting a new one at frame 0). Reconstruct the derived clock instead when runtime is behind elapsed. Adds a regression test.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -72,7 +72,17 @@ impl GgrsTimePlugin {
         // By scaling to nanoseconds, rounding error should be minimised.
         let runtime = Duration::from_nanos(this_frame * 1_000_000_000 / framerate);
 
-        time.advance_to(runtime);
+        // `GgrsTime` is fully derived from the frame count, so it must be able to
+        // follow the frame count backwards. For example, when a session is stopped
+        // and a new one started, the frame count resets to 0 while this clock still
+        // holds the previous session's elapsed time. `advance_to` panics on backward
+        // movement, so reconstruct the clock in that case instead.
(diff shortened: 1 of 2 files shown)


johanhelsing:
We deliberately work around the assert here: https://github.com/bevyengine/bevy/blob/v0.18.1/crates/bevy_time/src/time.rs#L244-L251 I'm not sure if it was added simply because moving time backwards is problematic, or if it's just unexpected.

gschup (approved):

src/time.rs
```diff
@@ -72,7 +72,17 @@ impl GgrsTimePlugin {
         // By scaling to nanoseconds, rounding error should be minimised.
         let runtime = Duration::from_nanos(this_frame * 1_000_000_000 / framerate);
 
-        time.advance_to(runtime);
+        // `GgrsTime` is fully derived from the frame count, so it must be able to
+        // follow the frame count backwards. For example, when a session is stopped
+        // and a new one started, the frame count resets to 0 while this clock still
+        // holds the previous session's elapsed time. `advance_to` panics on backward
+        // movement, so reconstruct the clock in that case instead.
+        if runtime >= time.elapsed() {
```
gschup: the only comment I have: this if-statements reads in a misdirecting fashion: "If the time goes backwards, do x else y"

The comment explains it, but it would of course be nicer if the code would read like that, too. (e.g. `if session_restarted...`). 

Don't think that this is a blocker, though.

PR #132, opened 2026-06-20
Bevy 0.19

johanhelsing:
Relatively straightforward and simple port to 0.19.

I also tried to unify the resource vs component paths. E.g. ComponentSnapshotPlugin vs. ResourceSnapshotPlugin, since Resources are now just regular Components internally... However I ran into issues about how to mark them for rollback or not. On entities we add a RollbackId, but on resources? We would need to have observers or something and it could get messy. Maybe it's possible, but in any case I think a simple port is a good starting point.

Confirmed box_test_p2p was working on localhost


gschup (approved):

johanhelsing:
@gschup could we do a release as well?

